PR #4, opened 2019-02-07
Some install instructions

DarthHater:
Well, it sure would be nice if we tell people how to install `nancy` or build it themselves

* Changes to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,9 @@ At current time you have a few options:
 * Run `go get -u github.com/sonatype-nexus-community/nancy`
 * Nancy should now be available whereever your GOPATH is set
 * Run `dep ensure` in the root of the project
+* In the root of the project `go test ./...`
+* If tests checkout go ahead and run `go build`.
+* Use that binary where ever your heart so desires!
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,12 +29,32 @@ This project is called `nancy` as like the great detective herself, it looks for
...
+
+### Build from source
+
+* Run `go get -u github.com/sonatype-nexus-community/nancy`
+* Nancy should now be available whereever your GOPATH is set
+* Run `dep ensure` in the root of the project
+* In the root of the project `go test ./...`
+* If tests checkout go ahead and run `go build`.
+* Use that binary where ever your heart so desires!
+
```
bhamail: Since things also work with "go modules" is it worth mentioning that build approach too? e.g.

For the adventurous, we have `go.mod` files that enable you to build using [go modules](https://github.com/golang/go/wiki/Modules).
```
$ export GO111MODULE=on
$ go test ./...
$ go build
```

DarthHater: Not a bad idea at all. I was sidestepping it, but why not!

DarthHater: In: 0fca1ea

PR #6, opened 2019-02-08
Improve conversion to PURL

ken-duck:
Improve package conversion to PURL

* Update convertGopkgNameToPurl to generate better PURLs

### DIFF
--- a/packages/packages.go
+++ b/packages/packages.go
@@ -14,8 +14,12 @@
 package packages
 
 import (
-	"strings"
+	"regexp"
 )
+
+var githubPattern = regexp.MustCompile("^github.com/([^/]+)/([^/]+).*")
+var gopkg1Pattern = regexp.MustCompile("^gopkg.in/([^.]+).*")
+var gopkg2Pattern = regexp.MustCompile("^gopkg.in/([^/]+)/([^.]+).*")
 
 // Packages is meant to be implemented for any package format such as dep, go mod, etc..
 type Packages interface {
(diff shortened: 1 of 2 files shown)


packages/packages.go
```diff
@@ -28,12 +32,22 @@ type Packages interface {
...
+	case githubPattern.MatchString(name):
+		// Currently OSS Index's github format support is based on repository
+		// owner/name, so restrict the PURL name path.
+		//
+		// Once golang format support is improved we can switch to that and get
+		// more precise and refined results on repository sub-paths
+		name = githubPattern.ReplaceAllString(name, "github/$1/$2")
+
+	case gopkg2Pattern.MatchString(name):
+		name = gopkg2Pattern.ReplaceAllString(name, "github/$1/$2")
```
DarthHater: QUESTION: Not knowing enough here, but this uses the gopk2Pattern, which is: `("^gopkg.in/([^/]+)/([^.]+).*")`, should it be stuffing `github` in the replace?

packages/packages.go
```diff
@@ -28,12 +32,22 @@ type Packages interface {
...
+
+	case gopkg1Pattern.MatchString(name):
+		name = gopkg1Pattern.ReplaceAllString(name, "github/go-$1/$1")
```
DarthHater: QUESTION: Not knowing enough here, but this uses the gopkg1Pattern, which is: `"^gopkg.in/([^.]+).*"`, should it be stuffing `github` in the replace?

packages/packages.go
ken-duck: gopkg.in is a service that maps shortened URLs with github repositories following a certain pattern, so we are just following the same pattern to find the real URLs.

packages/packages.go
ken-duck: Same as above.

For some additional flavouring, a "proper" golang PURL for github looks like this: `pkg:golang/github.com/<owner>/<name>@<version>`

However, OSS Index has better support for github PURLs than golang PURLs at the moment, though improvements are in progress. For that reason we use github PURLs instead: `pkg:github/<owner>/<name>@<version>`

PR #7, opened 2019-02-11
Ossindex tests

bhamail:
add some tests of the ossindex code


ossindex/ossindex_test.go
```diff
@@ -0,0 +1,110 @@
...
+	coordinates, err := AuditPackages([]string{"nonexistent-purl"})
+	assert.Equal(t, []types.Coordinate(nil), coordinates)
+	assert.Equal(t, "[400 Bad Request] error accessing OSS Index", err.Error())
+}
+
+func TestAuditPackages_NewPackage(t *testing.T) {
+	teardownTestCase := setupTestCaseMoveCacheDb(t)
+	defer teardownTestCase(t)
+
+	purl := "pkg:github/BurntSushi/toml@0.3.1"
```
DarthHater: TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

ossindex/ossindex_test.go
```diff
@@ -0,0 +1,110 @@
...
+	coordinates, err := AuditPackages([]string{purl})
+
+	lowerCasePurl := strings.ToLower(purl)
+	expectedCoordinate := types.Coordinate{
+		Coordinates:     lowerCasePurl,
+		Reference:       "https://ossindex.sonatype.org/component/" + lowerCasePurl,
```
DarthHater: TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

ossindex/ossindex_test.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	assert.Equal(t, []types.Coordinate{expectedCoordinate}, coordinates)
+	assert.Nil(t, err)
+}
+
+func TestAuditPackages_SinglePackage_Cached(t *testing.T) {
+	teardownTestCase := setupTestCaseMoveCacheDb(t)
+	defer teardownTestCase(t)
+
+	purl := "pkg:github/BurntSushi/toml@0.3.1"
```
DarthHater: TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

ossindex/ossindex_test.go
```diff
@@ -0,0 +1,110 @@
...
+	// call twice to ensure second call always finds package in local cache
+	coordinates, err := AuditPackages([]string{purl})
+	assert.Nil(t, err)
+	coordinates, err = AuditPackages([]string{purl})
+
+	lowerCasePurl := strings.ToLower(purl)
+	expectedCoordinate := types.Coordinate{
+		Coordinates:     lowerCasePurl,
+		Reference:       "https://ossindex.sonatype.org/component/" + lowerCasePurl,
```
DarthHater: TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

ossindex/ossindex_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"github.com/sonatype-nexus-community/nancy/types"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"strings"
+	"testing"
+)
+
+func exists(filePath string) (exists bool) {
```
DarthHater: NICE!

ossindex/ossindex_test.go
```diff
@@ -0,0 +1,110 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Definitions and functions for processing the OSS Index Feed
+package ossindex
+
+import (
+	"github.com/sonatype-nexus-community/nancy/types"
+	"github.com/stretchr/testify/assert"
```
DarthHater: The golang community is kinda torn on whether to use asserts or not (https://golang.org/doc/faq#assertions). You don't have to change anything, as like they even concede they are super useful. Figured I'd share so you can be in on the drama as well, if you weren't :)

bhamail: > The golang community is kinda torn on whether to use asserts or not (https://golang.org/doc/faq#assertions). You don't have to change anything, as like they even concede they are super useful. Figured I'd share so you can be in on the drama as well, if you weren't :)

Interesting. I was not aware of the drama. ;) I guess I equate "assertions" to things like Java [assert](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html) (e.g. something that ships in productions code) rather than unit test "[assert](http://junit.sourceforge.net/javadoc/org/junit/Assert.html).xxx" (something that exists only in test code).

DarthHater: Yeah I think for tests it's totally fine, just interesting they didn't build it into code itself with an `assert` keyword or something akin.

ossindex/ossindex_test.go
bhamail: > TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

Yes! 11ce0bd, 32ffc12

ossindex/ossindex_test.go
bhamail: > TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

Yes! 11ce0bd, 32ffc12

ossindex/ossindex_test.go
bhamail: > TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

Yes! 11ce0bd, 32ffc12

ossindex/ossindex_test.go
bhamail: > TRIVIAL: Since this is used more than once you could extract it to a `const` or something akin for the package

Yes! 11ce0bd, 32ffc12

ossindex/ossindex_test.go
bhamail: :) Still feels a little odd that this is not part of some standard package, but then again where do you stop if you starting adding such things...

bhamail:
> These tests are a good start but it's likely you should use: https://golang.org/pkg/net/http/httptest/ to mock the call out to OSS Index, so that we aren't doing real calls out to the remote.

I've pushed changes to no longer call out to the "real" ossindex service.

Issue #9, opened 2019-02-14
400 bad request when checking Gopkg.lock


fitzoh:
Shared on behalf of @Tanner from gophers.slack.com

After installing via `go get sonatype-nexus-community/nancy/...` , the following error is received when attempting to scan the Gopkg.lock file:
```
badger 2019/02/13 16:03:11 INFO: All 0 tables opened in 0s
badger 2019/02/13 16:03:12 WARNING: While forcing compaction on level 0: Unable to fill tables
Error auditing packages - error: [400 Bad Request] error accessing OSS Index
```

Sample file from another user that also fails w/ a 400:
[Gopkg.lock.txt](https://github.com/sonatype-nexus-community/nancy/files/2863100/Gopkg.lock.txt)

(had to add a `.txt` extension so GitHub would accept the file upload)

fitzoh:
After adding an extra println...
```
			body, _ := ioutil.ReadAll(resp.Body)
			fmt.Println(string(body))
			return nil, errors.New("[" + resp.Status + "] error accessing OSS Index")
```

`{"code":400,"message":"Unable to process JSON"}`

fitzoh:
On a hunch I switched to the `package-purl` branch from #6 and the error goes away.

DarthHater:
This should be addressed now @fitzoh , @ken-duck merged #6 !

fitzoh:
Validated that the error is no longer present for the sample file, closing

PR #10, opened 2019-02-14
Print nancy version on error and set a default version of "development".

fitzoh:
Print nancy version on error and set a default version of "development".

Inspired by #9


fitzoh:
Added in 3eb60950cd49cf82c92c28a2772b674146efc29b @bhamail

PR #11, opened 2019-02-15
Add in a user-agent header 

DarthHater:
Add in a user-agent header on the request so that we can start seeing what version people end up using, if people are using `nancy` versus just hitting OSS Index on their own, etc...

* Add's a `User-agent` header to the request, sets it to `nancy-client/version`
* Abstracts the request code into it's own function (make the world a better place!)

### DIFF
--- a/ossindex/ossindex.go
+++ b/ossindex/ossindex.go
@@ -169,10 +169,12 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
 }
 
 func setupRequest(jsonStr []byte) (req *http.Request, err error) {
-	if req, err = http.NewRequest(
+	req, err = http.NewRequest(
 		"POST",
 		getOssIndexUrl(),
-		bytes.NewBuffer(jsonStr)); err != nil {
+		bytes.NewBuffer(jsonStr),
+	)
+	if err != nil {
 		return nil, err
 	}
 
--- a/ossindex/ossindex.go
+++ b/ossindex/ossindex.go
@@ -176,7 +176,7 @@ func setupRequest(jsonStr []byte) (req *http.Request, err error) {
 		return nil, err
 	}
 
-	req.Header.Set("User-agent", fmt.Sprintf("nancy-client/%s", buildversion.BuildVersion))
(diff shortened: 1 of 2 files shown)


ossindex/ossindex.go
```diff
@@ -170,3 +167,17 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
...
+
+func setupRequest(jsonStr []byte) (req *http.Request, err error) {
+	if req, err = http.NewRequest(
+		"POST",
+		getOssIndexUrl(),
+		bytes.NewBuffer(jsonStr)); err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("User-agent", fmt.Sprintf("nancy-client/%s", buildversion.BuildVersion))
```
fitzoh: Capitalize the `a`?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

ossindex/ossindex.go
```diff
@@ -170,3 +167,17 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
 	}
 	return results, nil
 }
+
+func setupRequest(jsonStr []byte) (req *http.Request, err error) {
+	if req, err = http.NewRequest(
+		"POST",
+		getOssIndexUrl(),
+		bytes.NewBuffer(jsonStr)); err != nil {
```
fitzoh: This is nitpicky and it was there already, but I'm really not a fan of the inline if for something that spans multiple lines, took me a second to realize where the request builder ended and it went back to the `if`

ossindex/ossindex.go
DarthHater: Done in: 2ae546a

ossindex/ossindex.go
DarthHater: Done in: 72e7291

ossindex/ossindex.go
DarthHater: Something kinda interesting, is `req.Header.Get("User-Agent")` and `req.Header.Get("User-agent")` both work despite changing case. This caught me off guard when I adjusted the test and was like whoa why does it still work. Just figured I'd point this out, as golang seems to ignore case on headers, likely due to people being weirdos and using all kinds of different cases in the wild?

PR #25, opened 2019-09-04
Add a -quiet flag

packrat386:
Some projects will have a very large number of dependencies installed,
and the default behaviour prints a message for each package whether or
not it has vulnerabilities. This adds a -quiet flag that would print
output for packages only if they have vulnerabilities, to make it
easier to see at a glance what issues exist. The summary information,
packages analyzed vs vulnerable, is always printed.

cc @bhamail / @DarthHater

### DIFF
--- a/audit/auditlog.go
+++ b/audit/auditlog.go
@@ -71,7 +71,7 @@ func logVulnerablePackage(noColor bool, idx int, packageCount int, coordinate ty
 
 // LogResults will given a number of expected results and the results themselves, log the
 // results.
-func LogResults(noColor bool, packageCount int, coordinates []types.Coordinate) int {
+func LogResults(noColor bool, quiet bool, packageCount int, coordinates []types.Coordinate) int {
 	vulnerableCount := 0
 
 	for i := 0; i < len(coordinates); i++ {
@@ -79,7 +79,9 @@ func LogResults(noColor bool, packageCount int, coordinates []types.Coordinate) 
 		idx := i + 1
 
 		if len(coordinate.Vulnerabilities) == 0 {
-			logPackage(noColor, idx, packageCount, coordinate)
+			if !quiet {
+				logPackage(noColor, idx, packageCount, coordinate)
+			}
 		} else {
 			logVulnerablePackage(noColor, idx, packageCount, coordinate)
 			vulnerableCount++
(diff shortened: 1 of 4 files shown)


packrat386:
Thanks, I'll get on that. Also do y'all know what's up with this travis failure? 

```
--- FAIL: TestAuditPackages_ErrorHttpRequest (1.05s)

    ossindex_test.go:121: 

        	Error Trace:	ossindex_test.go:121

        	Error:      	Not equal: 

        	            	expected: "invalid character \"\\\\\" in host name"

        	            	actual  : "invalid port \":46783\\\\\" after host"

        	            	

        	            	Diff:

        	            	--- Expected

        	            	+++ Actual

        	            	@@ -1 +1 @@

        	            	-invalid character "\\" in host name

        	            	+invalid port ":46783\\" after host

        	Test:       	TestAuditPackages_ErrorHttpRequest
```

I can't recreate it locally (it passes) and it seems like it's due to a discrepancy in the error message.

packrat386:
Ok so it turned out https://github.com/golang/go/commit/61bb56ad63992a3199acc55b2537c8355ef887b6 in go 1.13 changes the error message you get back and breaks that test (which I can reproduce on master running go 1.13 locally). I figure it's best to just drop the assertion here https://github.com/sonatype-nexus-community/nancy/blob/master/ossindex/ossindex_test.go#L121 , since it's best to avoid checking exact error messages from the standard library anyway, and we already checked the type and opcode to know that it's a URL parse error. 

Would you prefer I send that in a separate PR, or add that here? Or is there some other way you'd like to address that?

bhamail:
@packrat386 I'm fine with you solution of dropping that assertion, and I'm fine with adding the fix to this PR. Thanks!

packrat386:
@DarthHater || @ken-duck || @bhamail added the test fix, lmk if there's anything else you think the PR needs.

DarthHater:
LGTM! Thanks for doing this! Hopefully it helps you!

DarthHater:
Bit posthumous, but we tend to do this, add yourself to: https://github.com/sonatype-nexus-community/nancy/blob/master/CONTRIBUTORS.md once you get a chance, we love external contributors!

DarthHater:
Thought about this a lil late, but you should send us a PR with adding this to the README, etc... so people know about it!

packrat386:
Sure, see #26

PR #26, opened 2019-09-06
Add usage section to README

packrat386:
Add usage section to README

cc @bhamail / @DarthHater

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,24 @@
 
 `nancy` is a tool to check for vulnerabilities in your Golang dependencies, powered by [Sonatype OSS Index](https://ossindex.sonatype.org/).
 
-## Usage
+### Usage
 
-To use `nancy`, assuming you have a built version of it:
+```
+ ~ > nancy
+Usage:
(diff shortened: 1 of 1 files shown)


packrat386:
@fitzoh @DarthHater how does this look?

fitzoh:
Oops, left a comment earlier but it looks like my phone ate it.

Looks good to me, just missing the flag that was added in #30.

DarthHater:
Yeah, just needs the `-exclude-vulnerability` flag added @packrat386 !

packrat386:
Updated.

Y'all may want to consider standardizing on `lisp-case` vs `camelCase` for flags, but that can be done some other time.

fitzoh:
Yeah, that's a little gross when it's laid out like that

zendern:
Im happy to open a PR changing it. The only one that is not `lisp-case` or all lowercase is `noColor`. I'd throw a vote in for `lisp-case` because from the looks of it most other golang based clis (kubectl, packer, aptly, etc) are using `lisp-case`.

PR #27, opened 2019-09-08
Use dep classes/parser directly

### LINKED ISSUE #13: Use dep interfaces for parsing Gopkg.toml, etc...

As pointed out by @sdboyer, there are interfaces that `dep` exposes that we can use to parse things, rather than parsing it ourselves.

Nothing explicitly, but may as well use these, they exist and can help us keep on track with `dep` development in the future

Take a look at `dep` and find these, and replace the parsing we do with them.

* **Anything else?**
FUN!

zendern:
* Switches main.go to just call the dep parser and uses those classes to now build the list of purls
* Regarding this change, I could have probably buried this call to dep in parse.go and then returned the same ProjectList/Projects classes but wasn't sure it was worth the extra level of indirection but I'm open to changing it.  
* The dep parser requires GOPATH to be set so I pass this in based in the os env variable. I don't think this should be an issue but is something that is different. 

**Important thing I noticed while building this**
I created an integration test to see what the current code returned from a purls perspective and ended up with 6 results. But once I used dep directly it gave me 14. From the looks of it the previous code was only returning versions if they matched semver anything else was thrown away. It wasn't clear where this was happening so I have left it for now just passing all version no matter what. 

* Fixes #13 

cc @bhamail / @DarthHater


packages/dep.go
```diff
@@ -14,38 +14,25 @@
...
 
-// Dep is an implementation of Packages interface
-type Dep struct {
-	ProjectList types.ProjectList
-	GopkgPath   string
-}
-
 // ExtractPurlsFromManifest will convert Gopkg projects to Package URLs
-func (d Dep) ExtractPurlsFromManifest() []string {
+func ExtractPurlsUsingDep(project dep.Project) []string {
```
fitzoh: Comment is out of date, and the `UsingDep` suffix could probably be dropped (I think the type covers that sufficiently now).

main.go
```diff
@@ -67,21 +68,32 @@ func main() {
...
-			checkOSSIndex(purls, packageCount)
+		workingDir := slices[0]
+		if(workingDir == "Gopkg.lock"){
+			workingDir, _ = os.Getwd()
+		}
+		getenv := os.Getenv("GOPATH")
+		ctx := dep.Ctx{
+			WorkingDir: workingDir,
+			GOPATH:     getenv,
+			GOPATHs:    []string{getenv},
```
fitzoh: Is the duplicate GOPATH necessary?

https://godoc.org/github.com/golang/dep#Ctx

zendern: Nope. Looks like I can get rid of GOPATH. I started with that one and it failed with this error. I then added GOPATHs with success but didn't realize GOPATH wasn't needed at all. So I'll clean it up. 

```
=== RUN   TestExtractPurlsFromManifestUsingDep
Gopkg.lock
Gopkg.toml
--- FAIL: TestExtractPurlsFromManifestUsingDep (0.01s)
    dep_int_test.go:39: /Users/zender/dev/projects/oss/nancy/packages/dir047508631/src/projectname is not within a known GOPATH/src
FAIL```

File: main.go [hunk 3]
```diff
@@ -67,21 +68,32 @@ func main() {
 func doCheckExistenceAndParse() {
 	switch {
 	case strings.Contains(path, "Gopkg.lock"):
-		dep := packages.Dep{}
-		dep.GopkgPath = path
-		if dep.CheckExistenceOfManifest() {
-			dep.ProjectList, _ = parse.GopkgLock(path)
-			var purls = processPackages(dep)
-			var packageCount = len(purls)
+		slices := strings.Split(path, "/Gopkg.lock")
```
fitzoh: I think you might be able to do `filepath.Dir(path)`

https://golang.org/pkg/path/filepath/#Dir

main.go
fitzoh: Hmm... I wonder if it depends on where you run `nancy`

```
    GOPATH         string        // Selected Go path, containing WorkingDir.
    GOPATHs        []string      // Other Go paths.
```

zendern: Not sure.....here is what i just tried

Install a build that only sets GOPATH and does not set GOPATHs. Fails with an error.
image


Use that same build but try to run it from a different directory. Same error
image



Install a new build that only sets GOPATHs and does not set GOPATH. Works with no issues from either location. 
image
image

packages/dep_int_test.go
```diff
@@ -0,0 +1,116 @@
+// Copyright 2018 Sonatype Inc.
```
DarthHater: Technically, you own the copyright on this file, it's all you. It's a test so lol no one cares, but figured I'd point it out.

DarthHater:
@zendern ok so a quick one on dep, the reason we didn't include some of the results is because we can't scan them without good coordinates. The hash or master that you see in some (not semver) is because those we couldn't identify if it was good or bad. @ken-duck might be able to comment more, and maybe we can just include them in the list and say something a bit more about them (cannot scan due to non semver coordinate).

packages/dep_int_test.go
zendern: looking over it now. It looks like all the other tests have the same copyright on the file so I'm going to leave it

zendern:
> @zendern ok so a quick one on dep, the reason we didn't include some of the results is because we can't scan them without good coordinates. The hash or master that you see in some (not semver) is because those we couldn't identify if it was good or bad. @ken-duck might be able to comment more, and maybe we can just include them in the list and say something a bit more about them (cannot scan due to non semver coordinate).

Thats good to know. 

What I can do is add code to filter them out from the dep results. But it might be useful to know that a dep couldn't be scanned and print some form of warning. I guess the question would be do you think that would go on the Sonatype scanner side or on the nancy side?? 

Right now if i just send them as is it says `No known vulnerabilities against package/version` but nothing about the fact that it could actually do any validation because that version ins't semver.
image

DarthHater:
I agree it would be helpful to know it can't be scanned, I think, otherwise it adds a false sense of security. I think we can do that easier on the Nancy side than the OSS Index side, an idea is just a regex test on the version to see if it's semver (this should be isolated to `dep` as we do allow a non traditional version with `go mod` because we were able to use it as a coordinate), and then some sort of helpful message about them @zendern

zendern:
@DarthHater I did the thing to add a warning when something is not using semver. Instead of a regex is used the semver library that was already in the project dependencies to just let it do the logic of parsing the version. FWIW It was a transitive dependency when `dep` itself as added. Here is a screenshot of what it looks like by default. With --noColor and --quiet. 

image

zendern:
Rebased with master to resolve conflicts. Thus the force-push.

zendern:
@DarthHater same with this one. All cleaned up and rebased against master. Want to take another look??

testdata/dep/Gopkg.toml
```diff
@@ -0,0 +1,28 @@
...
+  version = "1.0.0"
+
+[[constraint]]
+  name = "github.com/jmank88/nuts"
+  version = "0.3.0"
+
+[prune]
+  non-go = true
+  go-tests = true
+  unused-packages = true
```
DarthHater: Trivial: newline (and very trivial, since this is auto generated)

PR #28, opened 2019-09-10
Initial work for CVE exclude list flag

### LINKED ISSUE #16: Add exclusions
I would like to add nancy to the build of a Go project and have it fail if a vulnerability is discovered.
To do this I would need an option to exclude certain vulnerabilities from the scan as there might be false positives and vulnerabilities that do not apply in a certain context.
This would be similar to e.g. the ossindex maven plugin: https://sonatype.github.io/ossindex-maven/maven-plugin/excludes.

DarthHater:
This PR aims to provide a way to provide a list of CVEs that you want to mark as "safe" essentially, and not cause Nancy to fail. 

* New type that implements `flag.Value`
* Implementation of `exclude` flag

After finishing this PR, one would be able to run `./nancy -exclude CVE123,CVE456` and have it ignore those vulnerabilities (likely by telling you they were ignored, too). 

* Fixes #16 

cc @bhamail / @DarthHater

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -14,6 +14,9 @@
 package types
 
 import (
+	"fmt"
+	"strings"
+
 	decimal "github.com/shopspring/decimal"
 )
 
@@ -43,4 +46,21 @@ type Projects struct {
 }
 type ProjectList struct {
 	Projects []Projects
+}
+
+type CveListFlag struct {
+	cves []string
+}
+
+func (cve *CveListFlag) String() string {
+	return fmt.Sprint(cve.cves)
+}
+
+func (cve *CveListFlag) Set(value string) error {
+	if len(cve.cves) > 0 {
+		return fmt.Errorf("The CVE Exclude Flag is already set")
+	}
+	cve.cves = strings.Split(value, ",")
+
+	return nil
 }
--- a/main.go
+++ b/main.go
@@ -16,19 +16,22 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
+	"strings"
+
 	"github.com/sonatype-nexus-community/nancy/audit"
 	"github.com/sonatype-nexus-community/nancy/buildversion"
 	"github.com/sonatype-nexus-community/nancy/customerrors"
 	"github.com/sonatype-nexus-community/nancy/ossindex"
 	"github.com/sonatype-nexus-community/nancy/packages"
 	"github.com/sonatype-nexus-community/nancy/parse"
-	"os"
-	"strings"
+	"github.com/sonatype-nexus-community/nancy/types"
 )
 
 var noColorPtr *bool
 var quietPtr *bool
 var path string
+var cveList types.CveListFlag
 
 func main() {
 	args := os.Args[1:]
@@ -36,6 +39,7 @@ func main() {
 	noColorPtr = flag.Bool("noColor", false, "indicate output should not be colorized")
 	quietPtr = flag.Bool("quiet", false, "indicate output should contain only packages with vulnerabilities")
 	version := flag.Bool("version", false, "prints current nancy version")
+	flag.Var(&cveList, "exclude", "Comma seperated list of CVEs to exclude")
 
 	flag.Usage = func() {
 		_, _ = fmt.Fprintf(os.Stderr, "Usage: \nnancy [options] </path/to/Gopkg.lock>\nnancy [options] </path/to/go.sum>\n\nOptions:\n")


fitzoh:
#16 points to the [maven plugin](https://sonatype.github.io/ossindex-maven/maven-plugin/excludes/) which can exclude both dependencies and specific vulnerabilities.

Should we change the flag to `-exclude-vulnerability` to leave the door open for that in the future?

DarthHater:
@fitzoh that's totally fine by me.

fitzoh:
One other thought, if we wanted to change things up and use posix flags there's a built in array type: https://godoc.org/github.com/spf13/pflag#StringArray

DarthHater:
@fitzoh interesting! I did it this way because lol stdlib or whatever, but I'm not attached to it. 

Other feedback in: a77e90c

ken-duck:
No pressure, just wondering if this could work against non-CVE vulnerabilities? Perhaps by using the UUID or something?

breml:
Just an other idea: 
When the list of CVE, which are needed to be excluded gets longer it would be nice to have this list in an external file (e.g. one CVE per line) with a flag `-exclude-vulnerability-from`, which takes a file as argument. This would also be helpful to manage the list of excluded CVE in a single place, if for example nancy is executed from multiple places (like Makefile, CI config file, etc.).

DarthHater:
@breml , great idea. I probably won't tackle that in this PR but I'm sure someone can stepping stone to it from this work.

DarthHater:
@ken-duck so far what I've done is `strings.Contains` so that you can use CVE, CWE, blah blah blah. The ID's from OSS Index would be cool to use I suppose but then it becomes a bit of vendor lock-in, and the IDs mean nothing to someone looking at them?

DarthHater:
Minus fixing some tests, this works now:

`./nancy -exclude-vulnerability CWE-89,CWE-79,CVE-2018-15178,CVE-2018-15192,CVE-2017-1000070,CVE-2019-11576,CVE-2017-11480 /go/src/github.com/sonatype-nexus-community/intentionally-vulnerable-golang-project/Gopkg.lock`

You can see it in action by running that command (assuming you cloned that other project). Take a gander!

audit/auditlog.go
```diff
@@ -33,7 +34,7 @@ func logPackage(noColor bool, idx int, packageCount int, coordinate types.Coordi
 	}
 }
 
-func logVulnerablePackage(noColor bool, idx int, packageCount int, coordinate types.Coordinate) {
+func logVulnerablePackage(noColor bool, idx int, packageCount int, coordinate types.Coordinate) (vulnerableCount bool) {
```
fitzoh: `vulnerableCount` is a bool?
Also it doesn't look like the return value is used?

fitzoh:
Oops, meant to push https://github.com/sonatype-nexus-community/nancy/pull/28/commits/497cbf93579d849bd32088d12ea9ed5255af59ae to a branch and PR in, sorry bout that

types/types.go
```diff
@@ -25,6 +28,16 @@ type Vulnerability struct {
 	CvssVector  string
 	Cve         string
 	Reference   string
+	Excluded    bool
+}
+
+//Mark the given vulnerability as excluded if it appears in the exclusion list
+func (v *Vulnerability) maybeExcludeVulnerability(exclusions []string) {
+	for _, ex := range exclusions {
+		if strings.Contains(v.Title, ex) {
```
fitzoh: I was a little concerned about the `strings.Contains`...
Is there a reason to not do an equality check?

Is there a risk of over-excluding if one vuln is a substring of an excluded vuln?

DarthHater: I used `string.Contains` because the field is generally: `[CVE-2018-18925] Gogs 0.11.66 allows remote code execution because it does not properly validate ...`, `CWE-601: URL Redirection to Untrusted Site ('Open Redirect')` or something akin to that. The CVE field only populates CVEs, this allows someone to exclude more at the tinyish risk that it might be overzealous. I don't think a title would have more than one ID in it's name. An alternative to this is allowing people to exclude by OSS Index ID, but that seems to require a working knowledge of OSS Index rather than CVEs/CWEs, etc... and it also looked like vendor lockin.

fitzoh: Would it make sense to do an equality check against the `cve`/`cwe`?

fitzoh: >The CVE field only populates CVEs, this allows someone to exclude more at the tinyish risk that it might be overzealous.

it looks like OSSIndex returns CWEs as well, just need to attempt to deserialize them?

> I don't think a title would have more than one ID in it's name.
From your example above, if I exclude ` CWE-89` don't I also exclude `CWE-890`?

DarthHater: That is a very fair point. When I was scanning a project (the intentionally vulnerable one) I only saw CVE populated maybe 70% of the time. Open to ideas!

DarthHater: @ken-duck thoughts here?

fitzoh: Proposed changes in #30 , output looks the same with your earlier exclusion example (just need to remember to clear the cache)

DarthHater: I feel illuminated now!

audit/auditlog.go
DarthHater: Thanks for nixing this in: cdaffbe

audit/auditlog_test.go
```diff
@@ -50,7 +52,7 @@ func createVulnerabilities(num int) (vulnerabilities []types.Vulnerability) {
 }
 
 func createVulnerability() (vulnerability types.Vulnerability) {
-	vulnerability.Cve = "7.8"
+	vulnerability.Cve = "nerabil"
```
bhamail: @DarthHater FWIW, my use of the string "nerabil" in unit tests was intended for matching/filtering against the `Vulnerability.Title` structure element, which contained the string "Vulnerability". That may have been a mistake if the intent was to filter against Vulnerability.Cve element instead. I'm a little confused about that "nerabil" sub-string being used as the Cve element value.

DarthHater: Changed the values in 526f570

PR #30, opened 2019-09-11
switch to strict check for cve/cwe exclusion

fitzoh:
Add `Cwe` field to `Vulnerability` and do strict comparison against exclusion list.


DarthHater:
Well, shiver me timbers!

PR #35, opened 2019-09-23
CVE exclusion file

### LINKED ISSUE #31: Allow Vulnerability Exclusion via a file

It would be neat if someone could check in a file that has the CVE's, vulnerabilities, etc... that they want excluded from Nancy. It would run as if you were running `./nancy -exclude-vulnerability` but take a file path or pick one up by default

It's just an improved way of allowing someone to use Nancy on their project

You'd likely create a new command line flag for a file, determine a type of file and format (comma seperated list or new line separated list come to mind), and either load it by default if present, or allow someone to specify a path for it

* **Anything else?**

Have fun! You can look at #28 to get an idea of how to put things together.

cc @bhamail / @DarthHater

zendern:
Add the ability to exclude CVE using a file. If a file named `exclude_vulnerabilities` is defined in the directory that nancy is running it will be used. You also have the option to pass in the path of said exclude file by using cli flag `-exclude-vulnerability-file`. This also has the ability be used with the other cli flag of `exclude-vulnerability` since all the vulnerabilities are just added to a list.

As part of this change I refactored the flag parsing out of main since i need to add more logic to do the file processing. This also made the change easier to test. Only thing I felt like could be done is the `Configuration` struct that was created should that maybe go under `types` and not embedded in the `configuration/parse.go`??

* Fixes #31

cc @bhamail / @DarthHater

### DIFF
--- a/configuration/parse.go
+++ b/configuration/parse.go
@@ -1,11 +1,13 @@
 package configuration
 
 import (
+	"bufio"
 	"errors"
 	"flag"
 	"fmt"
 	"github.com/sonatype-nexus-community/nancy/types"
 	"os"
+	"strings"
 )
 
 type Configuration struct {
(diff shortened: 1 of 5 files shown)


configuration/parse.go
```diff
@@ -0,0 +1,80 @@
...
+
+func Parse(args []string) (Configuration, error) {
+	config := Configuration{}
+	var excludeVulnerabilityFilePath string
+
+	flag.BoolVar(&config.NoColor, "noColor", false, "indicate output should not be colorized")
+	flag.BoolVar(&config.Quiet, "quiet", false, "indicate output should contain only packages with vulnerabilities")
+	flag.BoolVar(&config.Version, "version", false, "prints current nancy version")
+	flag.Var(&config.CveList, "exclude-vulnerability", "Comma separated list of CVEs to exclude")
+	flag.StringVar(&excludeVulnerabilityFilePath, "exclude-vulnerability-file", "./exclude_vulnerabilities", "Path to a file containing newline separated CVEs to be excluded")
```
DarthHater: I feel like we could make this file have a default location and get picked up automagickally, wdyt?

configuration/parse_test.go
```diff
@@ -0,0 +1,85 @@
+package configuration
```
DarthHater: Oh my! Thank you so much for writing a new test!!!!

configuration/parse.go
zendern: `./exclude_vulnerabilities` is the default path. So wherever you are running nancy from. I dont know what a good default would be but im up for suggestions. It's easy enough to just change that default value. 

Are you thinking something like `~/.sonatype/cve-exclude_vulnerabilities`??

DarthHater: I am kinda thinking in the root of the project, like `.nancy-ignore` or something like that, and if you already have a sane default that's great.

zendern: .nancy-ignore feels reasonable. I'll switch it to that. Root of the project might be difficult since I don't have any context to where people are running it from thus why I went with cwd

zendern:
@DarthHater all cleaned up and rebased against master. Want to take another look??

DarthHater:
@zendern looks like there's a conflict now that #27 landed, one more trip around the sun please :)

zendern:
> @zendern looks like there's a conflict now that #27 landed, one more trip around the sun please :)

Alright should be ready to go now.

PR #39, opened 2019-09-28
Making other log calls quieter, if requested

### LINKED ISSUE #38: Nancy sometimes logs HTTP responses

Run Nancy on a project

```shell
$ nancy -quiet go.sum
2019/09/26 11:24:03 Response: &{Status:200 OK StatusCode:200 Proto:HTTP/2.0 ProtoMajor:2 ProtoMinor:0 Header:map[Content-Length:[664] Content-Type:[application/vnd.ossindex.component-report.v1+json] Date:[Thu, 26 Sep 2019 09:24:03 GMT] Set-Cookie:[AWSALB=BWff6DfuYnR9mHrXBvHS5C/o66uxqRRKutqrdMQ4hSgioBHC5C8jI3tCp4ziVoed
PtWWDGjxg5fUeeaHhfzwttovAlo0WpAwSqd8p++k2MsU7qHKu+WIxSpeSU2F; Expires=Thu, 03 Oct 2019 09:24:03 GMT; Path=/]] Body:{cs:0xc0059a6140} ContentLength:664 TransferEncoding:[] Close:false Uncompressed:false Trailer:map[] Request:0xc00011a200 TLS:0xc0059940b0}

Audited dependencies: 5, Vulnerable: 0
```

Running Nancy on a project without obtrusive log statements

Disable these log statements if not in development/debugging mode.

* **Anything else?**

The `go.sum` file I was checking was the following: https://github.com/mastertinner/clong/blob/master/go.sum

cc @bhamail / @DarthHater

zendern:
Simply makes the handful of other log statements we have go away if the quiet flag is passed. 

* Fixes #38

cc @bhamail / @DarthHater

### DIFF
--- a/ossindex/ossindex.go
+++ b/ossindex/ossindex.go
@@ -66,7 +66,7 @@ func openDb(dbDir string) (db *badger.DB, err error) {
 }
 
 // AuditPackages will given a list of Package URLs, run an OSS Index audit
-func AuditPackages(purls []string, quiet bool) ([]types.Coordinate, error) {
+func AuditPackages(purls []string) ([]types.Coordinate, error) {
 	dbDir := getDatabaseDirectory()
 	if err := os.MkdirAll(dbDir, os.ModePerm); err != nil {
 		return nil, err
@@ -78,9 +78,7 @@ func AuditPackages(purls []string, quiet bool) ([]types.Coordinate, error) {
 
 	defer func() {
 		if err := db.Close(); err != nil {
-			if !quiet {
-				log.Printf("error closing db: %s\n", err)
-			}
+			log.Printf("error closing db: %s\n", err)
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


ossindex/ossindex.go
```diff
@@ -78,7 +78,9 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
 
 	defer func() {
 		if err := db.Close(); err != nil {
-			log.Printf("error closing db: %s\n", err)
+			if !quiet {
```
DarthHater: Rather than doing a lot of `if !quiet` statements, what if we had a helper method that takes the `log.Printf` or whatever statement, and does the check in there? Then you could not have to pass `quiet` all over the place too?

zendern: @DarthHater Ok if checks removed in the latest commit. Realized that `log` has the ability to just set the output. So if i just set the output to be `Discard` it would get us the same result.

DarthHater:
@zendern small conflict here if you want to resolve we can merge this

zendern:
@DarthHater All fixed up and converted over to use the log statement to get rid of the quiet if check.

PR #40, opened 2019-09-28
Removing duplicate logs just for coloring.

zendern:
Aurora supports enabling/disabling colors which allows us to just write
a single message out instead of two

cc @bhamail / @DarthHater


DarthHater:
@zendern since I merged #41 I think this needs freshened up! Can you update and I'll get it in?

zendern:
@DarthHater All cleaned up.

PR #53, opened 2019-11-07
Print out Nancy version

### LINKED ISSUE #52: Display nancy version by default
Display the nancy version for easy debugging

I'm using nancy from jenkins and the github action, where it's not entirely clear what version of nancy is running.  I don't have an easy way to run `nancy version`, which makes it difficult to debug

By displaying the nancy version for a normal run unless quiet mode is enabled.

* **Anything else?**

cc @bhamail / @DarthHater

DarthHater:
Shows Nancy version by default when you run

* Prints `buildversion.BuildVersion` in `main.go` before doing any parsing, or auditing

* Fixes #52 

cc @bhamail / @DarthHater

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,10 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/golang/dep"
 	"github.com/sonatype-nexus-community/nancy/audit"
 	"github.com/sonatype-nexus-community/nancy/buildversion"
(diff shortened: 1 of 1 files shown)


DarthHater:
@fitzoh it's xmas in November

main.go
```diff
@@ -52,6 +53,8 @@ func main() {
 		os.Exit(0)
 	}
 
+	fmt.Println("Nancy version: " + buildversion.BuildVersion)
```
fitzoh: Should we skip this for quiet mode?

DarthHater: In: b826562

PR #56, opened 2019-11-20
Fix for CGO_ENABLED=0 some times not working on release artifacts

### LINKED ISSUE #32: nancy requires glibc
I'm trying to run nancy as part of my build chain which uses an alpine tag (`1.13.0-alpine3.10`) of the [official golang docker image](https://hub.docker.com/_/golang).

The nancy binary would not run. To reproduce, create the following Dockerfile:

```
FROM golang:1.13.0-alpine3.10

RUN wget -O- -q "https://github.com/sonatype-nexus-community/nancy/releases/download/v0.0.24/nancy-linux.amd64-v0.0.24" > /usr/bin/nancy\
    && chmod +x /usr/bin/nancy
```

And build & run it like this: `docker run -it --rm $(docker build -q .) /usr/bin/nancy `

The result will be `standard_init_linux.go:211: exec user process caused "no such file or directory"`.

This is due to nancy being built with `CGO_ENABLED=1` (the default setting) against glibc. Since alpine does not bring a glibc but musl as their libc implementation this doesn't work.

As far as I can see nancy does not need `CGO_ENABLED` so the fix is simple: disable it. I will open a PR...

/cc @bhamail / @DarthHater

zendern:
Fix CGO_ENABLED=0 artifact some times being release. 

See here
https://docs.travis-ci.com/user/environment-variables/#defining-multiple-variables-per-item

To have these not spin up multiple builds you must define them all on
the same line. When they are multiple builds it is a crapshoot as far as
which one gets released b/c its first one wins and that is not
consistent. See comment in #32 that shows screenshots of why build 33 worked and then build 34 did not work.

* Fixes #32

@bhamail / @DarthHater

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ install:
   - go mod download
 script:
   - go test ./... -v
+  - go env
   - GOOS=linux GOARCH=amd64 go build -ldflags="-X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildVersion=$VERSION' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildTime=$time' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildCommit=$TRAVIS_COMMIT'" -o nancy-linux.amd64-$VERSION
   - GOOS=darwin GOARCH=amd64 go build -ldflags="-X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildVersion=$VERSION' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildTime=$time' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildCommit=$TRAVIS_COMMIT'" -o nancy-darwin.amd64-$VERSION
   - GOOS=windows GOARCH=amd64 go build -ldflags="-X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildVersion=$VERSION' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildTime=$time' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildCommit=$TRAVIS_COMMIT'" -o nancy-windows.amd64-$VERSION.exe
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,15 +9,15 @@ install:
...
   - GOOS=linux GOARCH=amd64 go build -ldflags="-X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildVersion=$VERSION' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildTime=$time' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildCommit=$TRAVIS_COMMIT'" -o nancy-linux.amd64-$VERSION
   - GOOS=darwin GOARCH=amd64 go build -ldflags="-X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildVersion=$VERSION' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildTime=$time' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildCommit=$TRAVIS_COMMIT'" -o nancy-darwin.amd64-$VERSION
   - GOOS=windows GOARCH=amd64 go build -ldflags="-X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildVersion=$VERSION' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildTime=$time' -X 'github.com/sonatype-nexus-community/nancy/buildversion.BuildCommit=$TRAVIS_COMMIT'" -o nancy-windows.amd64-$VERSION.exe
   - ./nancy-linux.amd64-$VERSION Gopkg.lock
   - ./nancy-linux.amd64-$VERSION go.sum
   - go list -m all | ./nancy-linux.amd64-$VERSION
 env:
-  - GO111MODULE=on
-  - CGO_ENABLED=0
+  - GO111MODULE=on CGO_ENABLED=0
```
bhamail: Gah! Nice fix!

Prompted me to lookup this: https://docs.travis-ci.com/user/environment-variables/#defining-multiple-variables-per-item

PR #57, opened 2019-11-26
Multi format output

### LINKED ISSUE #45: [Feature request] output option in json format
Trying to figure out how to output in json format

exporting results in json makes parsing and understanding easy and helpful

have an argument like  `--json`

cc @bhamail / @DarthHater

zendern:
Adds the ability to output in different formats. Currently available are as follows: 

- text - same as it is today
- csv - a human readable csv output. Could be parsed by a machine but probably less likely. Does support the quiet option
- json - thats it a json representation of the output. Does not respect quiet and will look the same whether its passed or not.
- json-pretty - same as above but formatted so that a human could actually read it.

^^^ Examples of all the output can be found in the README.md

Used logrus to do the outputting and defined a formatter for each output type. If we ever find we want to add a new output type it should make it pretty easy to do. 

* Fixes #45 

cc @bhamail / @DarthHater


DarthHater:
@zendern quick question, how do I test this?

DarthHater:
NEVERMIND I'M A BAD READER.

zendern:
> NEVERMIND I'M A BAD READER.

It happens but still let me know if you have questions and we can get the docs in better shape if need be.

DarthHater:
Just did a quick test of this, and it's awesome. Dunno how much value someone will get from csv personally! XML would be a nice one to do as a follow up, but in the junit test case style (so we'd want a pretty custom reporter).

README.md
```diff
@@ -73,6 +77,536 @@ CVN-123 # Mitigated the risk of this since we only use one method in this packag
...
...tion of Input During Web Page Generation ('Cross-site Scripting')","Description":"The software does not neutralize or incorrectly neutralizes user-controllable input before it is placed in output that is used as a web page that is served to other users.","CvssScore":"6.1","CvssVector":"CVSS:3.0/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N","Cve":"","Reference":"https://ossindex.sonatype.org/vuln/fc70a115-52cc-44ea-a33d-793267f860dd","Excluded":false}],"InvalidSemVer":false},{"Coordinates":"pkg:golang/github.com/gophish/gophish@0.1.1","Reference":"https://ossindex.sonatype.org/component/pkg:golang/github.com/gophish/gophish@0.1.1","Vulnerabilities":[{"Id":"0416e202-2705-431d-9915-8ed93334ca58","Title":"CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')","Description":"The software does not neutralize or incorrectly neutralizes user-controllable input before it is placed in output that is used as a web page that is served to other users.","CvssScore":"6.1","CvssVector":"CVSS:3.0/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N","Cve":"","Reference":"https://ossindex.sonatype.org/vuln/0416e202-2705-431d-9915-8ed93334ca58","Excluded":false}],"InvalidSemVer":false}]}
```
DarthHater: You might wanna snip these down a bit more, they are looking pretty big in my review.

zendern:
> Just did a quick test of this, and it's awesome. Dunno how much value someone will get from csv personally! XML would be a nice one to do as a follow up, but in the junit test case style (so we'd want a pretty custom reporter).

re: XML - Yeah that would be cool to do as a followup. 
re: CSV - Yeah I kind of agree.....again thought was maybe it would be something you could easily ship to a security team member or something for easy "user-friendly" review. The content/format is 100% up for debate.

DarthHater:
I love what you did with outputting the exclusions, invalid, etc..., that's really great!

I'm going to need to sit down a tiny bit more with this tonight but I'm loving it so far.

README.md
zendern: Fixed up here 9af1cce

audit/jsonformatter.go
```diff
@@ -0,0 +1,27 @@
+package audit
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	. "github.com/sirupsen/logrus"
+)
+
+type JsonFormatter struct {
```
DarthHater: Curious, logrus has it's own JSONFormatter, why did you implement a new one? I'm reasonably sure I know the reason, just checking!

audit/jsonformatter.go
```diff
@@ -0,0 +1,27 @@
...
+	PrettyPrint bool
+}
+
+func (f *JsonFormatter) Format(entry *Entry) ([]byte, error) {
+	// Note this doesn't include Time, Level and Message which are available on
+	// the Entry. Consult `godoc` on information about those fields or read the
+	// source of the official loggers.
+	var b *bytes.Buffer = &bytes.Buffer{}
```
DarthHater: Pretty sure you can skip all the `bytes.Buffer` stuff and just use `serialized, err := json.Marshal(entry.Data)`, it returns a `[]byte, err`. You could just use https://golang.org/pkg/encoding/json/#MarshalIndent as well, rather than setting up the encoder, etc...

main.go
```diff
@@ -108,32 +106,33 @@ func doCheckExistenceAndParse() {
...
 	case strings.Contains(config.Path, "go.sum"):
 		mod := packages.Mod{}
 		mod.GoSumPath = config.Path
 		if mod.CheckExistenceOfManifest() {
 			mod.ProjectList, _ = parse.GoSum(config.Path)
 			var purls = mod.ExtractPurlsFromManifest()
-			var packageCount = len(purls)
 
-			checkOSSIndex(purls, packageCount)
+			checkOSSIndex(purls, []string{})
```
DarthHater: Pretty sure you can pass `nil` rather than instantiating `[]string`.

In the near future I'm going to likely move from `[]string` to `[]Purl`, as we can use this package to pass around our purls in a nicer way: https://github.com/package-url/packageurl-go, and we could add a `Valid bool` on to that as well, simplifying some of the code downstream.

audit/jsonformatter.go
zendern: Yeah so the comment I added here kind of alludes to it but its not 100% clear. But if i were to use the default json formatter we would end up with those fields as well. Honestly I did it because I felt like they didn't provide value and in the other more custom cases (csv and text) I wasn't going to include them either because they dont make a ton of sense in those outputs so I wanted to make it consistent across all of them at least. 

This does point out that we are kind of sort of abusing logrus to do this output. It's not bad but since we are ignoring fields its not standard if you will.

```
// Note this doesn't include Time, Level and Message which are available on
// the Entry. Consult `godoc` on information about those fields or read the
// source of the official loggers.
```

audit/jsonformatter.go
zendern: Good call.... that cleans things up quite a bit. See here a52348a

main.go
zendern: That sounds good....i still just put the nil slice in this commit b03d1aa

Nerzal:
This feature would be awesome, can we have this as fast as possible please? :D

DarthHater:
@Nerzal working on it!

@zendern you got a tiny conflict to fix up, if you want to clean that up I'll get this merged ASAP.

zendern:
@DarthHater all fixed up....should be good to go in a minute.

Issue #58, opened 2019-12-08
Missing Gopkg.lock causes crash


flimzy:
In the case that `Gopkg.toml` exists but `Gopkg.lock` does, not, nancy crashes with a nil pointer dereference:

```
Nancy version: development
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x935557]

goroutine 1 [running]:
github.com/sonatype-nexus-community/nancy/packages.ExtractPurlsUsingDep(0xc00002a124, 0x2b, 0xc00002b170, 0x2b, 0xc00002a139, 0x16, 0xc0001749a0, 0x0, 0x0, 0xc00002a139, ...)
        /home/jonhall/go/src/github.com/sonatype-nexus-community/nancy/packages/dep.go:23 +0x37
main.doCheckExistenceAndParse()
        /home/jonhall/go/src/github.com/sonatype-nexus-community/nancy/main.go:102 +0x214
main.main()
        /home/jonhall/go/src/github.com/sonatype-nexus-community/nancy/main.go:63 +0x104
```

To reproduce, create an empty directory containing a `Gopkg.toml` file, then attempt to run nancy.  Example:

```shell
$ mkdir asdf
$ cd asdf
asdf$ touch Gopkg.toml
asdf$ nancy Gopkg.lock
```

zendern:
So the scenario here would be that you have a project that is incorrectly setup (or is this a thing that does happen) and Nancy should instead of bombing at least tell you there is no lock file??

flimzy:
Yes, I think that's about right.  I don't know if it's ever legitimate to have `Gopkg.toml` without `Gopkg.lock` (probably not). But a proper error message seems like the correct behavior.

I'm not sure if this is a problem with nancy, though, or with the `dep` package. I spent a little time debugging, and it may be that `dep` just ignores that case, sending back an incomplete `package` variable to nancy.

zendern:
True... could be a dep “problem” but should be easy enough to guard against with a nil check and a better error message in Nancy.

PR #59, opened 2019-12-08
Avoid copying a lock by value, which is invalid.

flimzy:
It is not valid to copy a `sync.Mutex` value. From the [docs](https://golang.org/pkg/sync/#Mutex):

> A Mutex must not be copied after first use.

Doing so can lead to undefined results.

This PR fixes that, by passing a pointer to the value, rather than dereferencing the value before passing it.


flimzy:
Incidentally, I discovered this while debugging #58, thanks to running `golangci-lint` in my editor. I notice there are several other linter failures in this repo, some of which may represent real bugs. If you'd find it valuable, I'd be happy to submit fixes for all of them. It might be worth adding `golangci-lint` to your Travis-CI script as well.

fitzoh:
looks like `go vet` catches it as well:
```
$ go vet
# github.com/sonatype-nexus-community/nancy
./main.go:102:56: call of packages.ExtractPurlsUsingDep copies lock value: github.com/golang/dep.Project contains sync.Once contains sync.Mutex
```

flimzy:
> looks like go vet catches it as well:

`golangci-lint` calls `go vet` (and many other linters), so that's probably what I was running, too :)

zendern:
Well we are doing neither in CI and probably makes sense to do something there.

@flimzy what if we open an issue, get a little feedback and see about adding at least `go vet` if not `golangci-lint` to the project. 

@flimzy are you just running the default setup for `golangci-lint` or do you have other linters/checks enabled?? 

^^^I think the biggest thing with introducing that one is we just have to come to a little agreement what we do and don't want enabled. And really once we do it here it's probably going to get copied to other golang projects here like ahab.

flimzy:
> @flimzy what if we open an issue, get a little feedback and see about adding at least go vet if not golangci-lint to the project.

Makes sense to discuss in an issue, rather than here :)

> @flimzy are you just running the default setup for golangci-lint or do you have other linters/checks enabled??

I was using the default configuration. Although in most projects, I do use a custom config to enable a few non-default linters.

DarthHater:
@zendern @flimzy this is technically an API change (although I doubt it's going to really hurt anyone), should we bump to 0.1.0 for Nancy?

zendern:
> @zendern @flimzy this is technically an API change (although I doubt it's going to really hurt anyone), should we bump to 0.1.0 for Nancy?

Yeah i think that is reasonable to do.

DarthHater:
I'll try and get this merged tomorrow, I think it'll be Nancy 0.1.0 FTR! cc @zendern @flimzy

PR #63, opened 2019-12-09
Fix curl command in README

flimzy:
-O sets the remote name, -o is needed to set output location

cc @bhamail / @DarthHater


fitzoh:
Good catch, thanks!

PR #68, opened 2020-01-02
Nancy uses Nexus IQ Server, the journey

DarthHater:
This is a quick lil ditty to get Nancy to work with Nexus IQ Server

* Adds `iq.go` that handles communication with Nexus IQ
* Adds `cyclonedx` package that will turn a `[]packageurl.PackageURL` of purls into a minimal SBOM string
* Adds command line config options that shouldn't interfere with current OSS Index usage such that you call `./nancy iq -a application`

Successful output looks like:

```
(base) 685 me:nancy (NancyWithIQ *)$ go list -m all | ./nancy iq -application testapp
2020/01/02 12:09:12 Nancy version: development
..........
Wonderbar! No policy violations reported for this audit!
Report URL:  http://localhost:8070/ui/links/application/testapp/report/7a9655ecf31c40cb89149378b64e439a
```

Draft PR for now so we people can laugh at my implementation with less shame. I'll add README results after we sort what I did technically.

cc @bhamail / @zendern / @ken-duck / @ajbrown  / @fitzoh


go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/Masterminds/vcs v1.13.1 // indirect
 	github.com/armon/go-radix v1.0.0 // indirect
 	github.com/boltdb/bolt v1.3.1 // indirect
+	github.com/coreos/etcd v3.3.18+incompatible // indirect
```
DarthHater: I have zero clue where this came from, I'm guessing something upstream updated?

cyclonedx/cyclonedx.go
```diff
@@ -0,0 +1,70 @@
...
+type Components struct {
+	Component []Component `xml:"component"`
+}
+
+type Component struct {
+	Type    string `xml:"type,attr"`
+	BomRef  string `xml:"bom-ref,attr"`
+	Name    string `xml:"name"`
+	Version string `xml:"version"`
+	Purl    string `xml:"purl"`
```
DarthHater: I feel like all of this should go into `types.go`, any opinions?

configuration/parse.go
```diff
@@ -4,20 +4,27 @@ import (
...
+	Quiet       bool
+	Version     bool
+	CveList     types.CveListFlag
+	IQ          bool
+	Path        string
+	User        string
+	Token       string
+	Stage       string
+	Application string
+	Server      string
```
fitzoh: At risk of duplicating command line flags, it might make sense to add some docs here?

fitzoh: or... maybe an IqConfig struct instead of flattening everything?

configuration/parse.go
```diff
@@ -35,24 +42,53 @@ func Parse(args []string) (Configuration, error) {
 	flag.Var(&config.CveList, "exclude-vulnerability", "Comma separated list of CVEs to exclude")
 	flag.StringVar(&excludeVulnerabilityFilePath, "exclude-vulnerability-file", "./.nancy-ignore", "Path to a file containing newline separated CVEs to be excluded")
 
+	iqCommand := flag.NewFlagSet("iq", flag.ExitOnError)
+	iqCommand.StringVar(&config.User, "user", "admin", "Specify username for request")
+	iqCommand.StringVar(&config.Token, "token", "admin123", "Specify token/password for request")
```
fitzoh: those defaults feel kind of gross

configuration/parse.go
```diff
@@ -35,24 +42,53 @@ func Parse(args []string) (Configuration, error) {
...
+	iqCommand.StringVar(&config.Server, "server", "http://localhost:8070", "Specify Nexus IQ Server URL/port")
```
fitzoh: `server-url`?

configuration/parse.go
DarthHater: I can remove em, mostly just had em for like OOB use

configuration/parse.go
DarthHater: I like that! Config was completely new to me (I have been away too long!)

configuration/parse.go
fitzoh: Yeah, I get it...

Just can't shake the "encouraging people to use default user/pass" vibe

iq/iq.go
```diff
@@ -0,0 +1,191 @@
...
+	statusUrlResp = StatusURLResult{}
+
+	go func() {
+		for {
+			select {
+			case <-finished:
+				return
+			default:
+				pollIQServer(fmt.Sprintf("%s/%s", IQ_SERVER_BASE_URL, statusURL), finished)
+				time.Sleep(time.Second * 1)
```
fitzoh: should this be your `pollInterval`?

cyclonedx/cyclonedx.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		fmt.Print(err)
+	}
+
+	output = []byte(xml.Header + string(output))
+
+	return string(output)
+}
+
+func splitPurlIntoNameAndVersion(purl string) (name string, version string) {
```
fitzoh: This feels a little weird, but I'm not sure if it's worth fixing (would be nice if there was a Purl type you could just access these with)

DarthHater: I thought about that too. I think I'm going to do that because then people consuming it for other ecosystems can do the leg work and we won't need to do this.

iq/iq.go
DarthHater: In: 1126c52

cyclonedx/cyclonedx.go
fitzoh: have you looked at https://github.com/package-url/packageurl-go at all?

configuration/parse.go
DarthHater: In: 540319d

cyclonedx/cyclonedx.go
DarthHater: No! That rocks!

DarthHater: :boom: in 3803263

I kept it isolated to IQ for now so that I didn't have to touch any of the OSS Index stuff, but ideally we'd use it everywhere in the future.

configuration/parse.go
DarthHater: I'll probably write some helper warn function that if it sees defaults informs you that you are WILDIN out. WDYT?

fitzoh: I can get down with that.
Just want to make sure we're instilling a sense of shame in the user

fitzoh: >those defaults feel kind of gross

To clarify, it's really just the default password that feels gross

DarthHater: If someone is using `admin:admin123` I have now implemented this type of behavior:

```
(base) 756 me:nancy (NancyWithIQ *)$ go list -m all | ./nancy iq -application testapp --server-url http://localhost:8070
2020/01/02 13:28:29 Nancy version: development
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!! WARNING : You are using the default username and password for Nexus IQ, you should likely change these, and use a token. !!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
..........
Wonderbar! No policy violations reported for this audit!
Report URL:  http://localhost:8070/ui/links/application/testapp/report/23fa8b51f32e4ec1aec6b9a32dd888c5
```

DarthHater: Impl in: ea49c69

main.go
```diff
@@ -87,6 +93,23 @@ func doStdInAndParse() {
...
+		flag.Usage()
+		os.Exit(1)
+	} else {
+		mod := packages.Mod{}
+		scanner := bufio.NewScanner(os.Stdin)
+		mod.ProjectList, _ = parse.GoList(scanner)
+		var purls = mod.ExtractPurlsFromManifest()
+		auditWithIQServer(purls, config.Application, config)
+	}
+}
```
zendern: This duplication seems wrong to me. We should be able to simply check the config and switch out the audit methods. I think if you do it this way you will be missing out on some of the other stuff that is in `doCheckExistenceAndParse` that autodetects go.sum/Gopkg.lock for you. 

In your current state you will only be able to `|` in and do `go.mod` which maybe is fine ¯\_(ツ)_/¯ but its not consistent.

main.go
```diff
@@ -137,3 +160,23 @@ func checkOSSIndex(purls []string, packageCount int) {
 		os.Exit(count)
 	}
 }
+
+func auditWithIQServer(purls []string, applicationID string, config configuration.Configuration) {
```
zendern: The audit with IQ Server does not take into account any CVE exclusions that have been passed via the command line or define in .nancy-ignore file via the `config.CVEList`. Unless there is something I'm missing with Nexus IQ where that might not be wanted we should still support that feature. 

It might make sense to pull that up further. Right now we are doing it at the auditing level. Maybe we should pull that up into right after/during parsing of the purls so that all auditing code can take part in that without having to do it individually every time we add a new auditing mechanism.

main.go
```diff
@@ -62,9 +63,14 @@ func main() {
...
+	if config.UseStdIn && config.IQ {
+		doStdInAndParseForIQ(config)
+	}
+
+	if config.UseStdIn && !config.IQ {
 		doStdInAndParse()
-	} else {
+	}
+	if !config.UseStdIn && !config.IQ {
 		doCheckExistenceAndParse()
```
zendern: Same goes for these in that it feels weird. See comment for next block....feels like we could maybe push the config.IQ check down into just where we call audit and not need it here.

iq/iq.go
```diff
@@ -0,0 +1,221 @@
...
+		statusURLResp = response
+		if response.IsError {
+			finished <- true
+		}
+		finished <- true
+	}
+	fmt.Print(".")
+}
+
+func warnUserOfBadLifeChoices() {
```
zendern: best
#bestMethodName

iq/iq.go
```diff
@@ -0,0 +1,221 @@
...
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			fmt.Print(err)
```
zendern: If this fails should we exit??

iq/iq.go
```diff
@@ -0,0 +1,221 @@
...
+
+	req.Header.Set("Content-Type", "application/xml")
+	req.Header.Set("User-Agent", fmt.Sprintf("nancy-client/%s", buildversion.BuildVersion))
+
+	resp, err := client.Do(req)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
```
zendern: We do this alot..... we already have a `customerrors.Check` in error.go that basically does this and would clean up the if checks all over in this file. Thoughts on using that??

iq/iq.go
```diff
@@ -0,0 +1,221 @@
...
+	return
+}
+
+// AuditPackages accepts a slice of purls, public application ID, and configuration, and will submit these to
+// Nexus IQ Server for audit, and return a struct of StatusURLResult
+func AuditPackages(purls []string, applicationID string, config configuration.Configuration) StatusURLResult {
+	localConfig = config
+
+	if localConfig.User == "admin" && localConfig.Token == "admin123" {
```
zendern: Maybe pull "admin" and "admin123" out into a const to be used both here and in the Config class??

cyclonedx/cyclonedx.go
zendern: Yeah i think that would make sense

iq/iq.go
```diff
@@ -0,0 +1,221 @@
...
+	StatusURL string `json:"statusUrl"`
+}
+
+// StatusURLResult is a struct to let the consumer know what the response from Nexus IQ Server was
+type StatusURLResult struct {
+	PolicyAction  string `json:"policyAction"`
+	ReportHTMLURL string `json:"reportHtmlUrl"`
+	IsError       bool   `json:"isError"`
+	ErrorMessage  string `json:"errorMessage"`
+}
```
zendern: If we are moving things to `types.go` these should probably be moved as well.

configuration/parse.go
```diff
@@ -35,24 +42,53 @@ func Parse(args []string) (Configuration, error) {
...
-	if err != nil {
-		return config, err
+	if len(os.Args) > 1 {
+		if os.Args[1] == "iq" {
+			err := iqCommand.Parse(os.Args[2:])
+			if err != nil {
+				return config, err
+			}
+			config.IQ = true
+			config.UseStdIn = true
```
zendern: So this might answer some of my other questions on why you did it the way you did....you are forcing the users to only use the `|` option by doing this. Personally i feel like making the api consistent so that if im using OSS or IQ server it should work the same. Whether by pipe or passing in the file to be scanned. 

Really defining IQ is just saying i want you to use a different underlying audit infrastructure. Nancy is just a facilitator and shouldn't behave differently b/c of that switch out.

configuration/parse.go
```diff
@@ -35,24 +42,53 @@ func Parse(args []string) (Configuration, error) {
...
+	iqCommand.StringVar(&config.Server, "server-url", "http://localhost:8070", "Specify Nexus IQ Server URL/port")
+	iqCommand.StringVar(&config.Application, "application", "", "Specify application ID for request")
+	iqCommand.StringVar(&config.Stage, "stage", "develop", "Specify stage for application")
+
 	flag.Usage = func() {
-		_, _ = fmt.Fprintf(os.Stderr, "Usage: \nnancy [options] </path/to/Gopkg.lock>\nnancy [options] </path/to/go.sum>\n\nOptions:\n")
+		_, _ = fmt.Fprintf(os.Stderr, `Usage:
+	nancy [options] </path/to/Gopkg.lock>
+	nancy [options] </path/to/go.sum>
+	nancy iq [options]
```
zendern: This is out  of date in that technically we can do this with just `go list -m all | nancy [options]` as well :) not really your problem its just that you touched it so now I noticed.

README.md
```diff
@@ -7,22 +7,24 @@
...
 
 ### Usage
 
 ```
  ~ > nancy
 Usage:
-go list -m all | nancy [options]
-nancy [options] </path/to/Gopkg.lock>
-nancy [options] </path/to/go.sum>
+        nancy [options] </path/to/Gopkg.lock>
```
zendern: why did this option get deleted?? `go list -m all | nancy [options]`

main.go
```diff
@@ -137,3 +160,23 @@ func checkOSSIndex(purls []string, packageCount int) {
...
+		fmt.Println("There was an error with your request to Nexus IQ Server", res.ErrorMessage)
+		os.Exit(1)
+	}
+
+	if res.PolicyAction != "Failure" {
+		fmt.Println("Wonderbar! No policy violations reported for this audit!")
+		fmt.Println("Report URL: ", res.ReportHTMLURL)
+		os.Exit(0)
+	} else {
+		fmt.Println("Sonabot here, you have some policy violations to clean up!")
```
zendern: Why the reference to Sonabot?? nancy is not Sonabot so is that an IQ server reference??

README.md
```diff
@@ -73,6 +87,33 @@ CVN-123 # Mitigated the risk of this since we only use one method in this packag
...
+
+Successful submissions to Nexus IQ Server will result in either an OS exit of 0, meaning all is clear and a response akin to:
+
+```
+Wonderbar! No policy violations reported for this audit!
+Report URL:  http://reportURL
+```
+
+Failed submissions will either indicate failure because of an issue with processing the request, or a policy violation. Both will exit with a code of 1, allowing you to fail your build in CI. Policy Violation failures will include a report URL where you can learn more about why you encountered a failure.
+
```
zendern: You give an example of success message maybe add one for what an error message would look like as well ¯\_(ツ)_/¯

README.md
```diff
@@ -73,6 +87,33 @@ CVN-123 # Mitigated the risk of this since we only use one method in this packag
...
+A more logical use of `nancy` against Nexus IQ Server will look like so:
+
+`go list -m all | ./nancy iq -application public-application-id -user nondefaultuser -token yourtoken -server-url http://adifferentserverurl:port -stage develop`
+
+Options for stage are as follows:
+
+`build, develop, stage-release, release`
+
+By default `-stage` will be `develop`.
```
zendern: Why would the default be `develop`?? Im going to guess 99% of the time this will be running on CI so shouldn't it be `build`. Keep in mind i have no real idea what this stages mean in IQ (only briefly read some of the docs :) ) so I'll defer to you of course.

DarthHater: You have `build` > `develop` > `release-stage` > `release`. I went with develop as a default as Nancy can be used by developers locally pretty easy. We could set it to `build` for sure, but it kinda depends on how you setup your policy TBH with Nexus IQ. I originally had it as `build` ftr. I can go either way!

README.md
DarthHater: I've yet to get an error, lolololol.

zendern:
I assume you saw the tests are broken....looks like you can't run it via passing go.sum or Gopkg.lock anymore ¯\_(ツ)_/¯ 

Also tests are a lacking in this PR but probably wouldn't hurt to add at least something around the cyclonedx stuff and maybe something around the api calls into IQ

README.md
DarthHater: In: 6839696

DarthHater:
@zendern yeah, I saw all that, gonna fix all that up tomorrow, just wanted to get some early feedback on my approach!

main.go
DarthHater: We had an account back in the day called "Sonabot", I used that phrase in the work we did on `jake` and `auditjs` mostly to be consistent but let me switch it to `nancy` :)

README.md
DarthHater: GOOD QUESTION.

main.go
DarthHater: In: 772d557

iq/iq.go
DarthHater: LOL I knew someone would like that

main.go
DarthHater: Right, so with Nexus IQ Server you wouldn't manage any of the CVE exclusions, etc... inside of `nancy` or your repo, that's all managed in Nexus IQ Server, including policy based on what you want to cause failures, waivers on what you think is something you can live with, etc...

README.md
zendern: Cool cool.... that makes sense then. 

#scopeCreepAndMaybeNotWorthIt
But would it make sense to detect if on CI (looking for TRAVIS, CIRCLECI, CI, etc etc) environment variables and if one of those is set you change the stage automagically to `build`/`release-stage`?? Maybe that is still back into the "how you setup your policy in Nexus IQ" again and it doesn't matter. ¯\_(ツ)_/¯

iq/iq.go
DarthHater: I THINK WE SHOULD USE IT, MOSTLY BECAUSE I GAVE YOU A HARD TIME FOR NOT USING IT :)

iq/iq.go
DarthHater: I was on the fence about that. Probably. I was just worried that you start polling and maybe there is a quick network blip? I actually wanted to just do a like "120 max tries" approach, or something like that.

configuration/parse.go
DarthHater: Yeah, I want to make it work for `dep` but not parsing `go.sum` (since `go list -m all` is the preferred way to do this, more authoritative)

configuration/parse.go
DarthHater: In: https://github.com/sonatype-nexus-community/nancy/pull/68/commits/e46bd59d868878314a2cdcf1cff10ec1f144558f

README.md
DarthHater: I LIKE THAT. We'd still need to allow someone to override it, though, because people use Travis, CircleCI etc... to do CD too.

cyclonedx/cyclonedx.go
DarthHater: I'll handle that in the AM.

iq/iq.go
fitzoh: beautiful

packages/mod.go
```diff
@@ -37,6 +37,18 @@ func (m Mod) ExtractPurlsFromManifest() (purls []string) {
 	return purls
 }
 
+func (m Mod) ExtractPurlsFromManifestForIQ() (purls []string) {
+	for _, s := range m.ProjectList.Projects {
+		if len(s.Version) > 0 { // There must be a version we can use
+			var version = s.Version
+			version = strings.Replace(version, "+incompatible", "", -1)
```
zendern: So this and OSS index version of this method the only real different in what is being replaced. 

IQ = we replace "+incompatible" with ""
OSS = we repalce "v" with ""

Is it really true that each one can handle different version strings. Or is each on currently handling a different edge case that we might want for both??

iq/iq.go
zendern: So then swallow the error?? Or at least collect it and print it only if it truly fails?? Feels like noise if its really just a retry.

packages/mod.go
DarthHater: There is a healthy internal convo going about this here predicament, right now :)

Right now OSS Index prefers no `v` but I found a case today where it took one, which caused the conversation.

IQ demands a `v` at the moment.

I stripped `+incompatible` because it really isn't SemVer, afaik. I did however notice it didn't cause an issue with OSS Index, and I think it has to do with how OSS Index does version ranges.

configuration/parse.go
zendern: That's cool. Im going to open an issue to remove support for `go.sum` if we want to move away. No sense in supporting both.

README.md
zendern: Yeah for sure...if you pass it in then that is the value to be used. But nancy will attempt to "smart default" it for you.

DarthHater: Nancy is so clever!

configuration/parse.go
zendern: https://github.com/sonatype-nexus-community/nancy/issues/69

packages/mod.go
zendern: o neat well....
godspeed

go.mod
zendern: image

Fun story if you run `go mod tidy` that one and another disappears ... but we pick up something else ¯\_(ツ)_/¯

packages/mod.go
DarthHater: Outcome of the convo: strip the `v` for OSS Index, keep it for `IQ`. Pretty sure I reduce some of the code duplication by just doing a method that calls the other method with the data sanitized or something akin, OR we could create a new method that takes a function as a parameter, and then runs that function on the version. WDYT?

iq/iq.go
DarthHater: I went this route: https://github.com/sonatype-nexus-community/nancy/pull/68/commits/93eceafaf5ae40deea8e64c0c58ad36bc9be696e

go.mod
DarthHater: I ran tidy as well. Wonder why toml disappeared? I'd think dep would need that. Sometimes the mysteries of the world make zero sense to me.

main.go
DarthHater: I think I'm fine here, based on explanation about wanting to move away from using `go.sum`, resolve this?

cyclonedx/cyclonedx.go
DarthHater: Types I wanted public moved in: https://github.com/sonatype-nexus-community/nancy/pull/68/commits/4adf87f2bc642b09ff7b42208f1209a1b78cfa54

Made SBOM types public (because I'll eventually create a method to return a struct rather than a string, so people can do more with it).

main.go
DarthHater: That is totally a great point!

DarthHater:
@zendern tests fixed (and ability to specify path to go.sum, etc...) in: 4de41cc

DarthHater:
Errm maybe not! I'll take a gander.

DarthHater:
@zendern @fitzoh ok, tests pass, tear this thing apart :)

configuration/parse.go
DarthHater: Did an IQConfig in: https://github.com/sonatype-nexus-community/nancy/pull/68/commits/706eedc0682b1c5132999e07ee536aa9236ebc4f

main.go
DarthHater: I think I kinda dealt with this a bit better in: https://github.com/sonatype-nexus-community/nancy/pull/68/commits/706eedc0682b1c5132999e07ee536aa9236ebc4f

fitzoh:
>@zendern @fitzoh ok, tests pass, tear this thing apart :)

We were a little busy this past week :)
https://www.codemash.org/session-details/?id=145836

Might try to poke around a bit this weekend.

zendern:
So overall what you have here is good....my biggest concern (and it may not be valid since this is still a draft) but we still have no real test coverage around any of the Nexus IQ server stuff. 

It'd be good to see at least some unit tests around it but would be awesome if we were able to also add an integration tests like we have for our other public interfaces into nancy (go.sum, go.lock, go list, etc) that run as part of a build.  

To do that though we are looking at something like this to spin up an IQ server before we run anything.  
https://docs.travis-ci.com/user/docker/#using-a-docker-image-from-a-repository-in-a-build

I still dont know enough about IQ server to know if that is actually likely or not. I see a docker file here https://github.com/sonatype/docker-nexus-iq-server/blob/master/Dockerfile so hopefully ¯\_(ツ)_/¯

DarthHater:
I’m thinking I’ll likely setup a mock server that sends the responses back I expect, etc... for unit tests/integration tests. IQ Server needs a license etc... so doing a full fledged setup would be tricky. 

I’m on vacation till next week so I wouldn’t expect to see anything till then!

DarthHater:
@zendern , ok so I did a few things:

- Nancy now calls OSS Index before it calls IQ, the reason for this is to report on Go mod stuff that has the timestamp version hash thing, OSS Index currently has that data.
- Changed some of the stuff around so I could use OSS Index results to create a cyclonedx_bom
- Wrote a test! I swear! It's a happy path for now but it mocks out OSS Index, and a theoretical IQ Server, and returns the responses that you'd get in a happy world

Probably some more stuff. Would love feedback!

main.go
```diff
@@ -29,21 +29,31 @@ import (
 	"github.com/sonatype-nexus-community/nancy/buildversion"
 	"github.com/sonatype-nexus-community/nancy/configuration"
 	"github.com/sonatype-nexus-community/nancy/customerrors"
+	"github.com/sonatype-nexus-community/nancy/iq"
 	"github.com/sonatype-nexus-community/nancy/ossindex"
 	"github.com/sonatype-nexus-community/nancy/packages"
 	"github.com/sonatype-nexus-community/nancy/parse"
 )
 
-var config configuration.Configuration
```
DarthHater: I found it was easiest to just push around config rather than have it global given the new Config types

zendern:
> @zendern , ok so I did a few things:
> 
> * Nancy now calls OSS Index before it calls IQ, the reason for this is to report on Go mod stuff that has the timestamp version hash thing, OSS Index currently has that data.
> * Changed some of the stuff around so I could use OSS Index results to create a cyclonedx_bom
> * Wrote a test! I swear! It's a happy path for now but it mocks out OSS Index, and a theoretical IQ Server, and returns the responses that you'd get in a happy world
> 
> Probably some more stuff. Would love feedback!

image
Does this mean i've officially been knighted or something 👑👑👑 🤣 🤣 🤣 🤣

zendern:
So first things first..... I know we have all been there in trying to fix stuff that works locally and doesn't on CI but can we squash these or whichever set makes sense to squash together?? 

image

cyclonedx/cyclonedx_test.go
```diff
@@ -0,0 +1,56 @@
...
.../x/crypto@v0.0.0-20190308221718-c2843e01d9a2\">\n                          <v:id>CVE-123</v:id>\n                          <v:source name=\"ossindex\">\n                               <v:url>http://www.google.com</v:url>\n                          </v:source>\n                          <v:ratings>\n                               <v:rating>\n                                    <v:score>\n                                         <v:base>5.8</v:base>\n                                    </v:score>\n                                    <v:vector>WhatsYourVectorVictor</v:vector>\n                               </v:rating>\n                          </v:ratings>\n                          <v:description>Hello I am a CVE</v:description>\n                     </v:vulnerability>\n                </v:vulnerabilities>\n           </component>\n           <component type=\"library\" bom-ref=\"pkg:golang/github.com/go-yaml/yaml@v2.2.2\">\n                <name>yaml</name>\n                <version>v2.2.2</version>\n                <purl>pkg:golang/github.com/go-yaml/yaml@v2.2.2</purl>\n                <v:vulnerabilities></v:vulnerabilities>\n           </component>\n      </components>\n </bom>"
```
zendern: Using a multiline string here i think would help immensely in just understanding is here and cleanup all the `\n`.

iq/iq.go
```diff
@@ -0,0 +1,212 @@
...
+	statusURLResp = types.StatusURLResult{}
+
+	go func() {
+		for {
+			select {
+			case <-finished:
+				return
+			default:
+				pollIQServer(fmt.Sprintf("%s/%s", localConfig.Server, statusURL), finished)
+				time.Sleep(pollInterval)
```
zendern: So if i understand this code path correctly we will continually just poll IQ Server indefinitely?? Should we at least stop after `n` number of retries at least??

iq/iq_test.go
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+
+// Package iq has definitions and functions for processing golang purls with Nexus IQ Server
+package iq
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/jarcoal/httpmock"
```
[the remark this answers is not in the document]
zendern: 👍  Im cool with that.....lib looks real nice and easy to use. I dont remember other tests we have but if you know of spots maybe we create an issue to convert those over??

ossindex/ossindex.go
```diff
@@ -43,6 +44,12 @@ var (
 	ossIndexUrl string
 )
 
+// func ClearCache() {
+// 	db, err := openDb(getDatabaseDirectory())
+// 	customerrors.Check(err, "Error opening database")
+// 	defer db.Close()
+// }
+
```
zendern: We should kill this if we dont plan on bringing it back.

ossindex/ossindex.go
```diff
@@ -59,8 +60,13 @@ func getOssIndexUrl() string {
 
 func openDb(dbDir string) (db *badger.DB, err error) {
 	opts := badger.DefaultOptions
-	opts.Dir = dbDir + "/" + dbValueDirName
-	opts.ValueDir = dbDir + "/" + dbValueDirName
+	if flag.Lookup("test.v") == nil {
```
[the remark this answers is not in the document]
zendern: So this feels real icky to me....i get why its here but using the `test.v` flag doesn't feel like the right thing to trigger on. If i run the tests with `go test` and dont add `-v` this path is not going to get hit. 

Maybe adding a custom Environment variable would make sense and the IQ tests that need it can set that Environment variable before they run the tests that. 

More intrusive and less icky :) change would be to figure out a way to pass in the DB into OSSIndex that way the tests could mock it out or tweak the path or whatever instead of needing this if check.

ossindex/ossindex.go
```diff
@@ -74,14 +86,9 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
 
 	// Initialize the cache
 	db, err := openDb(dbDir)
+	defer db.Close()
```
zendern: I think we might want this after error check?? I'm guessing its possible to have an error returned and db to be nil. So this code will panic and not be a great user experience.

ossindex/ossindex.go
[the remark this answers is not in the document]
DarthHater: Yeah I didn't really dig this either, but I did like what it accomplished. I can try and remove the v?

I agree long term the best way is to inject the DB into ossindex, that's what I had hoped for to be honest.

iq/iq_test.go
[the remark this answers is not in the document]
DarthHater: Mostly just stuff for OSS Index. I'll create an issue once this PR hits master!

ossindex/ossindex.go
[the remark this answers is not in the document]
zendern: Yeah if we can remove the `v` that would be a little better since it wont rely on the test run needing to be in verbose mode.

cyclonedx/cyclonedx_test.go
DarthHater: I went another route, drug in etree and tested each element of the XML. This tests for validity for one, and for two ensures we've created what we want. In: 988416c

ossindex/ossindex.go
DarthHater: In: ce572ea , great catch!

ossindex/ossindex.go
DarthHater: In: 2e0f0fc

iq/iq.go
DarthHater: Did a bit in: 180312c

DarthHater:
@zendern I was going to squash and merge to master once I get there, do you want me to squash first?

ossindex/ossindex.go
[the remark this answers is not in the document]
DarthHater: In: ca93f03

zendern:
Re: squash into master. Nah that’s fine if you want to just do it that way. 👍

DarthHater:
@fitzoh @zendern marking this as ready, I think it's more or less ready.

README.md
DarthHater: I think we should create a ticket for this, it was a fantastic suggestion, however I dunno if I want to do it in this PR.

DarthHater:
Bit of a follow up to, I'm noticing it's hard to test error cases in Nancy because we throw around `os.Exit` like it's going out of style (a mistake of mine). Does anyone have objection to using `panic()` more often like I've been doing here? It makes it easier to test in some cases (ok my function panic'd, etc...)

cyclonedx/cyclonedx.go
DarthHater: Decided against struct for now because golang and xml unmarshall is actually kinda....bad! Loses some xmlns stuff, blew my mind. String was a way to guarantee nothing was lost.

packages/mod.go
DarthHater: Keeping this as it is for now.

zendern:
@DarthHater Latest changes look good to me. I do have a concern using panic and the fact that it now muddies user messaging when it is encountered. Basically it means two things....we lose control over the exit code coming back (maybe this is fine as long as it not 0 its fine) and we now give the user a stacktrace. Which sure nancy is a dev tool but feels like the wrong thing to show a user, most if any will not care about nancy internals. 

So few options come to mind....

1. Ignore me and just roll with it....dev can handle a stack trace
2. We leave the panic to make it easier to test but need to catch the panic at the top level and not present that stacktrace to the user. And maybe that would be nice b/c in the future we could add a flag to debug and it would spit out the stacktrace ¯\_(ツ)_/¯ 
3. Switch back to exit and change the test case to something like this. Where you spawn a subprocess test. 
https://stackoverflow.com/a/33404435

DarthHater:
@zendern I'm with you on the stack trace stuff. 

```
The convention in the Go libraries is that even when a package uses panic internally, its external API still presents explicit error return values.
```

I'm just trying to get more towards that, using panic in libs a bit more and allowing someone to recover if they want to, rather than explicitly force them to os.Exit

zendern:
@darthhater so then you plan is to put that into place?? So somewhere probably on main.go we capture the panic and return the message/correct error code

DarthHater:
Well, I'm kinda stuck on it personally. The middle ground is probably to panic, and then recover by throwing an error from our public APIs, that's consistent with the golang principle anyways. More so having an open conversation so I can figure out how to wrap this PR up, while heading some sort of direction :)

zendern:
@darthhater sooooo middle ground maybe to get this pushed along. The panic stuff you are testing error states right?? What if we go back to exit, scrap the test you wrote around error path and write up an issue to move the project from exit -> panic and implement the public error handler code that would be needed. Eat the tech debt around error handling testing in this case and get this shipped ¯\_(ツ)_/¯

DarthHater:
That sounds good, bud. I'll see what I can move forward today and then re-tag you in here!

DarthHater:
@zendern , moved back to errs and exits for now, if you wanna give this ole PR the rubber stamp!

PR #77, opened 2020-02-19
cli: add flag to delete local cache directory

adamdecaf:
* A quick way to delete the cache directory nancy uses. 

I didn't encounter a bug or corrupt cache, but wanted to delete the cache directory to make sure I was pulling the latest index. 

cc @bhamail / @DarthHater

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ nancy [options] </path/to/Gopkg.lock>
 nancy [options] </path/to/go.sum>
 
 Options:
+  -clean-cache
+    	Deletes local cache directory
   -exclude-vulnerability value
         Comma separated list of CVEs to exclude
   -exclude-vulnerability-file string
--- a/configuration/parse.go
+++ b/configuration/parse.go
@@ -4,20 +4,22 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
-	"github.com/sonatype-nexus-community/nancy/types"
 	"os"
 	"regexp"
 	"strings"
+
+	"github.com/sonatype-nexus-community/nancy/types"
 )
 
 type Configuration struct {
-	UseStdIn bool
-	Help bool
-	NoColor bool
-	Quiet bool
-	Version bool
-	CveList types.CveListFlag
-	Path    string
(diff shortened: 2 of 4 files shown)


ossindex/ossindex.go
```diff
@@ -50,6 +50,11 @@ func getDatabaseDirectory() (dbDir string) {
 	return usr.HomeDir + "/.ossindex"
 }
 
+// RemoveCacheDirectory deletes the local database directory.
+func RemoveCacheDirectory() error {
+	return os.RemoveAll(getDatabaseDirectory())
```
DarthHater: I'm wondering if there is a better way to do this (basically just flush Badger rather than delete the directory)

DarthHater: Wonderbar.

I've been wanting to move away from Badger for a lil bit (it's a bit overkill for what we need, to be honest). You've got a reasonable explanation for now! Thanks for answering!

README.md
```diff
@@ -19,6 +19,8 @@ nancy [options] </path/to/Gopkg.lock>
 nancy [options] </path/to/go.sum>
 
 Options:
+  -clean-cache
+    	Deleted local cache directory
```
DarthHater: Lil typo here, probably should be `Deletes local cache directory`

configuration/parse.go
```diff
@@ -32,6 +34,7 @@ func Parse(args []string) (Configuration, error) {
 	flag.BoolVar(&noColorDeprecated, "noColor", false, "indicate output should not be colorized (deprecated: please use no-color)")
 	flag.BoolVar(&config.Quiet, "quiet", false, "indicate output should contain only packages with vulnerabilities")
 	flag.BoolVar(&config.Version, "version", false, "prints current nancy version")
+	flag.BoolVar(&config.CleanCache, "clean-cache", false, "Deleted local cache directory")
```
DarthHater: Lil typo here, probably should be Deletes local cache directory

adamdecaf:
I've updated the `-clean-cache` doc line.

PR #84, opened 2020-02-28
Build scratch Docker image

ajbrown:
Changes the docker build to product an image from scratch, containing only the bare essentials for nancy.

* Dockerfile producing a from scratch image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest as builder
 
 RUN apk update \
     && apk upgrade \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest as builder
```
DarthHater: Do we really need to do a builder image? Is that to get the certs passed on to the scratch image? Not against it just trying to understand!

fitzoh: yeah, I believe so...
You could argue that `builder` is a bit misleading since we're not actually building it in the container?

Maybe `FROM alpine:latest as certs`?

Issue #88, opened 2020-03-09
Unintended behavioral change with StdIn, no config, and -quiet flag.


bhamail:
The latest version of Nancy is exiting with a 3 error code when invoked thusly:

     go list -m all | ../../nancy -quiet

This invocation used to work - it would parse the stdIn and produce results without an error status being returned.

Could be related to PR #78?

cc @bhamail / @DarthHater

bhamail:
Fixed by #89

PR #89, opened 2020-03-09
Fix parsing of CLI flags when using stdIn invocation.

bhamail:
Support passing in CLI flags when invoking Nancy using the standard-input pipe approach.

e.g.

```
go list -m all |  ./nancy -quiet
```

Fixes Issue #88 

cc @bhamail / @DarthHater


DarthHater:
@zendern you mind taking a gander here? The config stuff was largely your ball court, I think you'll have the best eye!

zendern:
@bhamail @DarthHater Found it and my bad!! 😢 
https://godoc.org/github.com/stretchr/testify/assert#ObjectsAreEqual

Apparently this does no assertion. Working on trying to switch it over to an method that does an assert but having issues with the logurus formatters since they are pointers.

zendern:
Now you should see a failing test for the standard in and quiet scenario in 38f8e29. Had to fiddle all the formatters.

bhamail:
Excellent. I'm off and running again. Thanks @zendern !

PR #94, opened 2020-03-12
Initial logging

### LINKED ISSUE #76: Implement logging for Nancy

Nancy could use some logging, for helping people diagnose issues like @fitzoh in https://github.com/sonatype-nexus-community/nancy/issues/71

It is just a way we can get some better detail from users when they run into issues

The suggestion is to:

- Log to `~/.ossindex`
- Provide a mechanism to specify the log location (if someone wants to log to `/tmp`, etc...
- Log debug by default to it's own file (all log levels), and then errors to an error file
- Maybe cycle logs on each request (so people don't end up with a growing log file and have no idea why)
- DO NOT LOG SENSITIVE INFO

* **Anything else?**
Have fun!?

cc @bhamail / @DarthHater

DarthHater:
Welp, it's about time we got better about logging!

* Adds a logger package, so you can obtain a logger from anywhere
* Partially implements the logger for the IQ path (thus far)

* Fixes #76 

### DIFF
--- a/configuration/parse.go
+++ b/configuration/parse.go
@@ -39,6 +39,9 @@ type Configuration struct {
 	CveList    types.CveListFlag
 	Path       string
 	Formatter  logrus.Formatter
+	Info       bool
+	Debug      bool
+	Trace      bool
 }
 
 type IqConfiguration struct {
@@ -50,6 +53,9 @@ type IqConfiguration struct {
 	Application string
 	Server      string
 	MaxRetries  int
+	Info        bool
+	Debug       bool
+	Trace       bool
 }
 
 var unixComments = regexp.MustCompile(`#.*$`)
@@ -57,6 +63,9 @@ var untilComment = regexp.MustCompile(`(until=)(.*)`)
 
 func ParseIQ(args []string) (config IqConfiguration, err error) {
 	iqCommand := flag.NewFlagSet("iq", flag.ExitOnError)
+	iqCommand.BoolVar(&config.Info, "v", false, "Set log level to Info")
+	iqCommand.BoolVar(&config.Debug, "vv", false, "Set log level to Debug")
+	iqCommand.BoolVar(&config.Trace, "vvv", false, "Set log level to Trace")
 	iqCommand.StringVar(&config.User, "user", "admin", "Specify username for request")
 	iqCommand.StringVar(&config.Token, "token", "admin123", "Specify token/password for request")
 	iqCommand.StringVar(&config.Server, "server-url", "http://localhost:8070", "Specify Nexus IQ Server URL/port")
@@ -99,6 +108,9 @@ func Parse(args []string) (Configuration, error) {
 	flag.BoolVar(&config.Quiet, "quiet", false, "indicate output should contain only packages with vulnerabilities")
 	flag.BoolVar(&config.Version, "version", false, "prints current nancy version")
 	flag.BoolVar(&config.CleanCache, "clean-cache", false, "Deletes local cache directory")
+	flag.BoolVar(&config.Info, "v", false, "Set log level to Info")
+	flag.BoolVar(&config.Debug, "vv", false, "Set log level to Debug")
+	flag.BoolVar(&config.Trace, "vvv", false, "Set log level to Trace")
 	flag.Var(&config.CveList, "exclude-vulnerability", "Comma separated list of CVEs to exclude")
 	flag.StringVar(&excludeVulnerabilityFilePath, "exclude-vulnerability-file", "./.nancy-ignore", "Path to a file containing newline separated CVEs to be excluded")
 	flag.StringVar(&outputFormat, "output", "text", "Styling for output format. "+fmt.Sprintf("%+q", reflect.ValueOf(outputFormats).MapKeys()))
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -27,8 +27,9 @@ import (
 // DefaultLogFile can be overriden to use a different file name for upstream consumers
 var DefaultLogFile = "nancy.combined.log"
 
-// Logger can be obtained from outside the package
-var Logger = logrus.New()
+// LogLady can be obtained from outside the package, the name is a reference to the brilliant
+// actress in Twin Peaks
+var LogLady = logrus.New()
 
 func init() {
 	file, err := os.OpenFile(getLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
@@ -36,14 +37,14 @@ func init() {
 		fmt.Print(err)
 	}
 
-	Logger.Out = file
-	Logger.Level = logrus.InfoLevel
-	Logger.Formatter = &logrus.JSONFormatter{}
+	LogLady.Out = file
+	LogLady.Level = logrus.InfoLevel
+	LogLady.Formatter = &logrus.JSONFormatter{}
 }
 
 // SetLogLevel is a helper function to set the logging level to something other than it's default
 func SetLogLevel(level logrus.Level) {
-	Logger.Level = level
+	LogLady.Level = level
 }
 
 func getLogFileLocation() (result string) {
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -21,6 +21,7 @@ import (
 	"path"
 
 	"github.com/sirupsen/logrus"
+	"github.com/sonatype-nexus-community/nancy/types"
 )
 
 // DefaultLogFile can be overriden to use a different file name for upstream consumers
@@ -47,6 +48,6 @@ func SetLogLevel(level logrus.Level) {
 
 func getLogFileLocation() (result string) {
 	result, _ = os.UserHomeDir()
-	result = path.Join(result, ".ossindex", DefaultLogFile)
+	result = path.Join(result, types.OssIndexDirName, DefaultLogFile)
 	return
 }
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,0 +1,33 @@
+package logger
+
+import (
+	"os"
+	"path"
+
+	"github.com/sirupsen/logrus"
+)
+
+var DefaultLogFile = "nancy.combined.log"
+
+var Logger *logrus.Logger
+
+func NewLogger() *logrus.Logger {
+	file, err := os.OpenFile(getLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		Logger.Fatal(err)
+	}
+
+	logger := &logrus.Logger{
+		Out:       file,
+		Level:     logrus.DebugLevel,
+		Formatter: &logrus.JSONFormatter{},
+	}
+	Logger = logger
+	return Logger
+}
+
+func getLogFileLocation() (result string) {
+	result, _ = os.UserHomeDir()
+	result = path.Join(result, ".ossindex", DefaultLogFile)
+	return
+}
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -35,9 +35,10 @@ func init() {
 	}
 
 	logger := &logrus.Logger{
-		Out:       file,
-		Level:     logrus.DebugLevel,
-		Formatter: &logrus.JSONFormatter{},
+		Out:          file,
+		Level:        logrus.DebugLevel,
+		Formatter:    &logrus.JSONFormatter{},
+		ReportCaller: true,
 	}
 	Logger = logger
 }
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -32,7 +32,7 @@ var DefaultLogFile = "nancy.combined.log"
 var LogLady = logrus.New()
 
 func init() {
-	file, err := os.OpenFile(getLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+	file, err := os.OpenFile(GetLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
 	if err != nil {
 		fmt.Print(err)
 	}
@@ -47,7 +47,8 @@ func SetLogLevel(level logrus.Level) {
 	LogLady.Level = level
 }
 
-func getLogFileLocation() (result string) {
+// GetLogFileLocation will return the location on disk of the log file
+func GetLogFileLocation() (result string) {
 	result, _ = os.UserHomeDir()
 	result = path.Join(result, types.OssIndexDirName, DefaultLogFile)
 	return
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -32,7 +32,7 @@ var DefaultLogFile = "nancy.combined.log"
 var LogLady = logrus.New()
 
 func init() {
-	if len(os.Args) > 1 && os.Args[1][:5] == "-test" {
+	if len(os.Args) > 1 && os.Args[1] != "iq" && os.Args[1][:5] == "-test" {
 		DefaultLogFile = "nancy.test.log"
 	}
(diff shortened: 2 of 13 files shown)


zendern:
image

logger/logger.go
```diff
@@ -0,0 +1,50 @@
...
+		Level:        logrus.DebugLevel,
+		Formatter:    &logrus.JSONFormatter{},
+		ReportCaller: true,
+	}
+	Logger = logger
+}
+
+func getLogFileLocation() (result string) {
+	result, _ = os.UserHomeDir()
+	result = path.Join(result, ".ossindex", DefaultLogFile)
```
zendern: Maybe we pull `.ossindex` out into a constant somewheres?? Not sure what else will go there but now we will have badger and the log file ¯\_(ツ)_/¯

logger/logger.go
```diff
@@ -0,0 +1,50 @@
...
+func init() {
+	file, err := os.OpenFile(getLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		Logger.Fatal(err)
+	}
+
+	logger := &logrus.Logger{
+		Out:          file,
+		Level:        logrus.DebugLevel,
+		Formatter:    &logrus.JSONFormatter{},
```
zendern: I assume this is going to be more for human consumption.... if im being honest line after line of json output is not super readable to me so I'd vote for switching this out

main.go
```diff
@@ -39,63 +41,81 @@ import (
...
+
 	if len(os.Args) > 1 && os.Args[1] == "iq" {
+		appLog.Debug("Nancy parsing config for IQ")
 		config, err := configuration.ParseIQ(os.Args[2:])
 		if err != nil {
 			flag.Usage()
 			os.Exit(1)
 		}
 		processIQConfig(config)
+		appLog.Debug("Nancy finished parsing config for IQ")
```
zendern: Do we want even more printed here as far as the actually config that was parsed goes??

main.go
```diff
@@ -39,63 +41,81 @@ import (
...
 	}
 }
 
 func printHeader(print bool) {
 	if print {
+		appLog.Debug("Attempting to print header")
 		figure.NewFigure("Nancy", "larry3d", true).Print()
 		figure.NewFigure("By Sonatype & Friends", "pepper", true).Print()
 		log.Println("Nancy version: " + buildversion.BuildVersion)
+		appLog.Debug("Finished printing header")
```
zendern: While we are hear might as well dump the current Nancy version into the log file as well so that you dont have to ask people for it you just have to ask for the log file.

ossindex/ossindex.go
```diff
@@ -139,7 +141,7 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
 
 			defer func() {
 				if err := resp.Body.Close(); err != nil {
-					log.Printf("error closing response body: %s\n", err)
+					appLog.Debugf("error closing response body: %s\n", err)
```
zendern: Did you not do appLog.Error here b/c we did not Exit??

logger/logger.go
```diff
@@ -0,0 +1,50 @@
...
+
+func init() {
+	file, err := os.OpenFile(getLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		Logger.Fatal(err)
+	}
+
+	logger := &logrus.Logger{
+		Out:          file,
+		Level:        logrus.DebugLevel,
```
zendern: I see you used trace else where but there is no switch here. Are you thinking of adding more flags that can be passed to up the level. Maybe a good ole `-vvv` like in curl

zendern:
Overall I like it and it looks good....just dropped some questions and thoughts throughout.

logger/logger.go
DarthHater: I kinda went with JSON because then you could consume it programmatically easier. Fine with using a TextFormatter, too!

main.go
DarthHater: I deliberately avoided printing config as it could have SENSITIVE info.

ossindex/ossindex.go
DarthHater: Oversight on my part! I can switch to error easily

logger/logger.go
zendern: up to you really.....in all honesty i cant see a usecase where consuming it programmatically would happen but if you have a usecase sure why not :shipit:

logger/logger.go
DarthHater: Yep! That's the idea. Maybe set the logger to Error by default, -v does Debug, -vv Info, -vvv Trace?

logger/logger.go
DarthHater: I don't, just used to planning for the worst. I'll switch to Text!

main.go
zendern: Is it worthwhile to redact the sensitive info?? Passwords, urls, etc as part of logging?? A logrus hook like this could help. 

https://github.com/whuang8/redactrus

logger/logger.go
zendern: -v == Info 
-vv == Debug
-vvv == Trace

Youve got Info and Debug flipped i think. But 👍

DarthHater: Hah, likely. I never understood the levels anyways to begin with :)

logger/logger.go
DarthHater: Ok I switched to Text and I hate it, BECAUSE when you use `WithFields` it doesn't really output the field with it. You know a way around this?

zendern: What does it give you?? Docs show it should look like this

DarthHater: I get effectively:

```
time="2020-03-12T18:04:45-08:00" level=debug msg="Successful in communicating with IQ Server" res="{None http://localhost:8070/ui/links/application/testapp/report/4ca4b65c2eff4e4ca922fa608c78bbd2 false }"
```
What I don't like is to get it to do what you see, I have to take a struct and you know, add a field for each field, rather than be lazy and log the whole thing as "res". It's totally me being lazy, but it also allows someone to add fields to a struct and they get logged by default?

zendern: ah i see....JsonFormatter it is :) #imAlsoLazy

DarthHater: Note this is because it's to file, not to console, but I imagine to console it wouldn't print the object fields.

main.go
DarthHater: Redactrus works only on named fields, not on an object that you pass to logrus, sadly. It's also 2 years old and looks to be not maintained. Might be fun to cook something like that up!

logger/logger.go
DarthHater: Really good point!

DarthHater: Done in e2712b6

DarthHater:
Marking this as ready for review. It doesn't have logging EVERYWHERE yet, but I also don't know if I want to balloon this PR anymore. We can create Issues for implementing logging in additional packages, etc...

@zendern @bhamail @fitzoh LET THE TEARING OF MY WORK APART BEGIN!

main.go
```diff
@@ -40,118 +42,209 @@ import (
...
+			"version":      buildversion.BuildVersion,
+		}).Info("Printing version information and exiting clean")
+
 		fmt.Println(buildversion.BuildVersion)
 		_, _ = fmt.Printf("build time: %s\n", buildversion.BuildTime)
 		_, _ = fmt.Printf("build commit: %s\n", buildversion.BuildCommit)
 		os.Exit(0)
 	}
 
+	if config.Info || config.Debug || config.Trace {
```
bhamail: minor: Why this outer `if`?

main.go
```diff
@@ -40,118 +42,209 @@ import (
...
```
bhamail: minor: Why this outer if?

Maybe refactor into shared method for both OSSI and IQ? (or not...)

main.go
DarthHater: No real reason, I can drop it!

main.go
DarthHater: Right now I can't do that too easy, config has two different structs, and no common interface. Gonna have to leave it like this until we simplify config a bit (something I'd like to do)

main.go
```diff
@@ -40,118 +42,205 @@ import (
...
 }
 
 func printHeader(print bool) {
 	if print {
+		Logger.Info("Attempting to print header")
 		figure.NewFigure("Nancy", "larry3d", true).Print()
 		figure.NewFigure("By Sonatype & Friends", "pepper", true).Print()
+
+		Logger.WithField("version", buildversion.BuildVersion).Info("Printing Nancy version")
 		log.Println("Nancy version: " + buildversion.BuildVersion)
```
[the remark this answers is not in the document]
zendern: Down further when config.Version is set we log all the info around version. Do we want to do that here too?

useragent/useragent.go
```diff
@@ -49,34 +51,43 @@ func GetUserAgent() string {
...
 		return getUserAgent(fmt.Sprintf("github-action %s", id), callTree)
 	}
 
+	Logger.Trace("Returning User Agent")
 	return getUserAgent("ci usage", callTree)
 }
 
 func getUserAgent(agent string, callTree string) string {
+	Logger.Trace("Obtaining parsed User Agent string")
 	return fmt.Sprintf("%s (%s; %s %s; %s)", getUserAgentBaseAndVersion(), agent, GOOS, GOARCH, callTree)
```
zendern: Maybe pull this out into a var before returning it and log it too...might be useful to have the full built useragent in the future??

configuration/parse.go
```diff
@@ -99,6 +108,9 @@ func Parse(args []string) (Configuration, error) {
 	flag.BoolVar(&config.Quiet, "quiet", false, "indicate output should contain only packages with vulnerabilities")
 	flag.BoolVar(&config.Version, "version", false, "prints current nancy version")
 	flag.BoolVar(&config.CleanCache, "clean-cache", false, "Deletes local cache directory")
+	flag.BoolVar(&config.Info, "v", false, "Set log level to Info")
+	flag.BoolVar(&config.Debug, "vv", false, "Set log level to Debug")
+	flag.BoolVar(&config.Trace, "vvv", false, "Set log level to Trace")
```
zendern: We have a bunch of testing around parsing these and the effects of them probably should add some around to make sure the v's are doing what we expect.

logger/logger.go
```diff
@@ -0,0 +1,53 @@
...
+
+func init() {
+	file, err := os.OpenFile(getLogFileLocation(), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		fmt.Print(err)
+	}
+
+	Logger.Out = file
+	Logger.Level = logrus.InfoLevel
+	Logger.Formatter = &logrus.JSONFormatter{}
```
zendern: Is it worthwhile to create a test around this to make sure its actually configured to output a file and put it in the `.ossindex` dir??

main.go
bhamail: Doh! That makes total sense.

useragent/useragent.go
DarthHater: Good point!

logger/logger.go
DarthHater: PROBABLY

useragent/useragent.go
DarthHater: In: 4dd0e60

logger/logger.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/sirupsen/logrus"
+	"github.com/sonatype-nexus-community/nancy/types"
+)
+
+// DefaultLogFile can be overriden to use a different file name for upstream consumers
+var DefaultLogFile = "nancy.combined.log"
+
+// Logger can be obtained from outside the package
+var Logger = logrus.New()
```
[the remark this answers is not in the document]
DarthHater: ApplicationLogger ?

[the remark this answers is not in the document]
zendern: Im up for whatever Logger just feels too generic is all. `ApplicationLogger`, `FileLogger`, `TriageLogger`, `NotConsoleLogger`, `FileThatIfYouLookAtItsProbablyBrokenLogger`

main.go
```diff
@@ -40,118 +42,205 @@ import (
...
 		doStdInAndParse(config)
 	}
 	if !config.UseStdIn {
+		Logger.Info("Parsing config for file based scan")
 		doCheckExistenceAndParse(config)
 	}
 }
 
 func processIQConfig(config configuration.IqConfiguration) {
+	// TODO: a lot of this code is a duplication of the OSS Index config, probably should extract some of it
```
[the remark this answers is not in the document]
DarthHater: 💥 https://github.com/sonatype-nexus-community/nancy/issues/95

logger/logger.go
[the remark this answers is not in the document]
DarthHater: I am going to call it `LogLady` because this project needs a good Twin Peaks easter egg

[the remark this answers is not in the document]
DarthHater: In: https://github.com/sonatype-nexus-community/nancy/pull/94/commits/b91aebdffbb42662a04d3e053c06fcd517b38668

main.go
DarthHater: In: a0b1b44

logger/logger.go
bhamail: > ....JsonFormatter it is

Agreed! Stick with what was working before I stirred the pot. Thanks for looking into it!

logger/logger.go
DarthHater: Test here: 78e6990

main.go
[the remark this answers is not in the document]
DarthHater: Honestly we probably don't need the `log.Println` here now, since I limit this whole section to being printed if we have `quiet`. WDYT?

configuration/parse.go
DarthHater: NOOOOOOOOO (ok I'll do it)

main.go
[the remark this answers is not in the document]
zendern: id we added it as part of #52 so i believe removing it will basically revert that change.

configuration/parse.go
DarthHater: In 3dce67a and e3a3fb1

main.go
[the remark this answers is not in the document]
DarthHater: What I meant is now:

```
	if print {
		LogLady.Info("Attempting to print header")
		figure.NewFigure("Nancy", "larry3d", true).Print()
		figure.NewFigure("By Sonatype & Friends", "pepper", true).Print()

		LogLady.WithField("version", buildversion.BuildVersion).Info("Printing Nancy version")
		log.Println("Nancy version: " + buildversion.BuildVersion)
		LogLady.Info("Finished printing header")
	}
```

It's surrounded in `print`, which I think means we can switch the l`og.Println` to `fmt.Println`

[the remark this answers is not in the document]
zendern: Yeah and looking a little more i think we can actually remove this whole chunk now. 

``` 
if config.Quiet {
		LogLady.Debug("Setting console log output to discard, quiet requested")
		log.SetOutput(ioutil.Discard)
	}
```

Basically if we need to log something "tracing" wise we will use `LogLady` and if it needs to go to the console it should probably be baked into the logrus formatters we have now that spit out Json, Text, Csv, etc. I'm thinking it might make sense to bake the header output into the auditlog.go then we will have all the "console" output in one place. But i digress. 

If youre cool with it flip that to a fmt.Println and wipe that if check. I'll probably create an issue to clean up the header stuff and move it.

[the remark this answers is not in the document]
DarthHater: Will do!

PR #96, opened 2020-03-16
Remove dupes from list, add tests for that

DarthHater:
A few users have been scanning go projects with LOTS of sub modules, and they've been running into issues because some of the projects have duplicate dependencies, and this fails validation with IQ Server. So, let's solve that!

* Add a removeDuplicates function to `parse.go`
* Add two test cases for checking that de-duplication is happening

This is the first time I've gone "Hey Golang, why don't you have a Set!". Been writing too much TypeScript :)


parse/gosumdupes.sum
```diff
@@ -0,0 +1,24 @@
+github.com/AndreasBriese/bbloom v0.0.0-20180913140656-343706a395b7 h1:PqzgE6kAMi81xWQA2QIVxjWkFHptGgC547vchpUbtFo=
```
bhamail: maybe move this file into a `testdata` folder?

DarthHater: In: 0502bd8

bhamail:
Maybe for a different PR, but I did a little testing, and it appears these tests no longer use `testdata/Gopkg.lock`.

PR #102, opened 2020-03-18
If in quiet mode, don't print the extra newline before the results.

### LINKED ISSUE #101: Extra newline printed in quiet mode

Use the quiet mode. When I run it, I see:

image

Where there's an extra odd newline. This newline makes sense when not in quiet mode, as a line will be printed for each dependency, so this helps distinguish the results. But in quiet mode, it just looks weird.

Quiet mode.

If in quiet mode, don't print the extra newline before the results.

* **Anything else?**

cc @bhamail / @DarthHater

bhamail:
* Fixes #101

cc @bhamail / @DarthHater

### DIFF
--- a/audit/auditlogtextformatter_test.go
+++ b/audit/auditlogtextformatter_test.go
@@ -3,6 +3,7 @@ package audit
 import (
 	"errors"
 	. "github.com/sirupsen/logrus"
+	"github.com/sonatype-nexus-community/nancy/types"
 	"github.com/stretchr/testify/assert"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


audit/auditlogtextformatter_test.go
```diff
@@ -21,3 +22,30 @@ func TestFormatterErrorsIfEntryNotValid(t *testing.T) {
 	assert.NotNil(t, e)
 	assert.Equal(t, errors.New("fields passed did not match the expected values for an audit log. You should probably look at setting the formatter to something else"), e)
 }
+
+func verifyFormatterSummaryLoudness(t *testing.T, quiet bool) {
```
DarthHater: LOL LOUDNESSS

PR #106, opened 2020-04-01
Check for existence of unmarshalled internal ID before returning

DarthHater:
Basically we need to check if the unmarshall was successful, when sending a public application ID to Nexus IQ Server

* Quick logic check before returning
* Add some logging around the check to help someone find the issue later

cc @bhamail / @DarthHater

### DIFF
--- a/iq/iq_test.go
+++ b/iq/iq_test.go
@@ -103,6 +103,26 @@ func TestAuditPackages(t *testing.T) {
 	assert.Equal(t, result, statusExpected)
 }
 
+func TestAuditPackagesIqCannotLocateApplicationID(t *testing.T) {
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	httpmock.RegisterResponder("GET", "http://sillyplace.com:8090/api/v2/applications?publicId=testapp",
+		httpmock.NewBytesResponder(200, []byte(`{ "applications": [] }`)))
+
+	var purls []string
+	purls = append(purls, "pkg:golang/github.com/go-yaml/yaml@v2.2.2")
+	purls = append(purls, "pkg:golang/golang.org/x/crypto@v0.0.0-20190308221718-c2843e01d9a2")
+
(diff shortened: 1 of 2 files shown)


iq/iq_test.go
```diff
@@ -103,6 +103,26 @@ func TestAuditPackages(t *testing.T) {
...
+	httpmock.RegisterResponder("GET", "http://sillyplace.com:8090/api/v2/applications?publicId=testapp",
+		httpmock.NewBytesResponder(200, []byte(`{ "applications": [] }`)))
+
+	var purls []string
+	purls = append(purls, "pkg:golang/github.com/go-yaml/yaml@v2.2.2")
+	purls = append(purls, "pkg:golang/golang.org/x/crypto@v0.0.0-20190308221718-c2843e01d9a2")
+
+	_, err := AuditPackages(purls, "testapp", setupIqConfiguration())
+	if err == nil {
+		t.Error("There is an error")
```
zendern, line 119: Huh?? Did you mean  `There must be an error` or something like that ?

iq/iq_test.go
```diff
@@ -103,6 +103,26 @@ func TestAuditPackages(t *testing.T) {
...
+	}
+	if err.Error() != "Internal ID for testapp could not be found, or Nexus IQ Server is down" {
```
zendern, line 121: So is the IQ server down really the case?? I should shouldnt this not really be an either or?? More of a definitive yes its up an the app isnt found and yes your IQ server is down and broken??

iq/iq_test.go
DarthHater: Typo from the original PR, updated to something better in: 2b127ad

iq/iq_test.go
DarthHater: I changed the function a bit to return better errors (and errors in general) in: 2b127ad

Now you should get errors that reflect a bit better what actually happened instead of an omnibus error (as well as logging around it).

PR #108, opened 2020-04-04
Get and Set Config for OSS Index, IQ Server

DarthHater:
Last week much to my chagrin I discovered a few things:

- Nancy currently has no way of specifying OSS Index config, which means it's hard to escape rate limiting if you run in to it!
- I never built the ability to use a persistent config for IQ Server!

* Adds `set.go` to `package configuration` and wires up to it in `main.go` for `nancy config`. Methods made public so that `cheque` and `ahab` can consume them easily as well
  * Config is saved to `~` and then `.ossindex` or `.iqserver`, and as `.oss-index-config` or `.iq-server-config` to be consistent with other tools out of the DevEx factory (set once, picked up by the other tools, great success)
* Adds an ability to loadConfigFromFile to `parse.go`, which we really should refactor at some point (there is so much duplication it's making me go ahhhhhhh, but not doing that in this PR quite yet)
* Chunks up `ossindex.go` so that I can pass in configuration a bit easier. Also decomposes some larger methods into smaller ones. Adds a new public function of `AuditPackageWithOSSIndex`, so that if someone is currently using `AuditPackages` upstream, they can still do that. Add's a deprecation notice to `AuditPackages` with a note to use the new function instead.
* Adds tests for saving config, and verifying it was saved properly

Couple lil fly by improvements:
* `LogLady` now truncates when it opens the log file, ensuring the log file is only ever for the most recent use of `nancy`, noticed this while helping on #107 
* Passes a better error when you are rate limited, which explains how to get an OSS Index account and set your config


configuration/parse.go
```diff
@@ -83,6 +91,16 @@ Options:
 		os.Exit(2)
 	}
 
+	usersHomeDir, _ := os.UserHomeDir()
+
+	path := filepath.Join(usersHomeDir, types.IQServerDirName, types.IQServerConfigFileName)
+
+	err = loadIQConfigFromFile(path, &config)
+	if err != nil {
+		fmt.Println(err)
```
DarthHater, line 100: Should fix this up

configuration/parse.go
```diff
@@ -91,6 +109,32 @@ Options:
 	return config, nil
 }
 
+func loadConfigFromFile(configLocation string, config *Configuration) error {
```
DarthHater, line 112: These two functions are essentially the same. One way to avoid this is to change the signature to allow to pass in a `config *interface{}` which would allow me to do what I am doing with one method, but it becomes INFINITELY reusable at that point. That should be fine since the function is private, but we could also just simplify config a lot to accomplish one function, by combining the configuration structs, and working from there.

ossindex/ossindex.go
```diff
@@ -86,8 +87,20 @@ func openDb(dbDir string) (db *badger.DB, err error) {
 	return
 }
 
-// AuditPackages will given a list of Package URLs, run an OSS Index audit
+// AuditPackages will given a list of Package URLs, run an OSS Index audit.
+//
+// Deprecated: AuditPackages is old and being maintained for upstream compatibility at the moment.
```
zendern, line 92: Maybe open an issue so that we can clean this up before we cut a 1.0.0 official release??

configuration/parse.go
```diff
@@ -128,7 +173,17 @@ Options:
 		os.Exit(2)
 	}
 
-	err := flag.CommandLine.Parse(args)
+	usersHomeDir, _ := os.UserHomeDir()
+
+	path := filepath.Join(usersHomeDir, types.OssIndexDirName, types.OssIndexConfigFileName)
+
```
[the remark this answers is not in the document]
DarthHater: I'm hesitant to allow this, because by using a common config location, the rest of the tools (`jake`, etc...) can pick up the settings and run with them. That makes it easy for someone to use all the tools, some of the tools, etc...

WDYT

configuration/set.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+func getAndSetIQConfig() {
+	iqConfig := IQConfig{Server: "http://localhost:8070", Username: "admin", Token: "admin123"}
+	fmt.Print("What is the address of your Nexus IQ Server (default: http://localhost:8070)? ")
+	fmt.Scanln(&iqConfig.Server)
+	fmt.Print("What username do you want to authenticate as (default: admin)? ")
+	fmt.Scanln(&iqConfig.Username)
+	fmt.Print("What token do you want to use (default: admin123)? ")
```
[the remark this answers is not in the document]
DarthHater: Yeah, which is why I defaulted it! I could also just warn someone if they use the defaults and go "DO YOU REALLY WANT TO DO THIS?" and if they say no, rerun the function and recapture them? WDYT?

ossindex/ossindex.go
```diff
@@ -189,23 +173,49 @@ func AuditPackages(purls []string) ([]types.Coordinate, error) {
...
 
-		if end > len(purls) {
-			end = len(purls)
+	if resp.StatusCode == http.StatusTooManyRequests {
+		LogLady.WithField("resp_status_code", resp.Status).Error("Error accessing OSS Index due to Rate Limiting")
+		return nil, &types.OSSIndexRateLimitError{}
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		LogLady.WithField("resp_status_code", resp.Status).Error("Error accessing OSS Index")
```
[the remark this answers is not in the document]
DarthHater: On a 400, absolutely. It's likely something we should do in general, handle error codes more "explicitly" rather than catch all like this.

ossindex/ossindex.go
```diff
@@ -218,6 +228,26 @@ func setupRequest(jsonStr []byte) (req *http.Request, err error) {
 
 	req.Header.Set("User-Agent", useragent.GetUserAgent())
 	req.Header.Set("Content-Type", "application/json")
+	if config != nil && config.Username != "" && config.Token != "" {
+		LogLady.Info("Set OSS Index Basic Auth")
+		req.SetBasicAuth(config.Username, config.Token)
```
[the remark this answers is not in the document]
DarthHater: They'd like be split out as the basic auth creds (so base64 encoded), so not the actual values. I can check on this to be sure. We could potentially look at something like this again, too: https://github.com/whuang8/redactrus

types/types.go
```diff
@@ -23,7 +23,10 @@ import (
 
 // Helpful constants to pull strings we use more than once out of code
 const (
-	OssIndexDirName = ".ossindex"
+	OssIndexDirName        = ".ossindex"
+	OssIndexConfigFileName = ".oss-index-config"
+	IQServerDirName        = ".iqserver"
+	IQServerConfigFileName = ".iq-server-config"
```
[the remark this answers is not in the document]
DarthHater: They use that same folder, so I've been in the habit of putting stuff there, so someone can easily locate ALL of the oss index tool logs, etc...

ossindex/ossindex.go
DarthHater: YES ZENDER SIR! 🖖

configuration/set.go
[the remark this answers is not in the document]
zendern: Yeah I'm cool with that

configuration/parse.go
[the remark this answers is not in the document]
zendern: Makes sense. 👍

configuration/set.go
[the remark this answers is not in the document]
DarthHater: How is this?

```
(base) 504 jeffryhesse:nancy (ConfigOptions *)$ ./nancy config
Hi! What config can I help you set, IQ or OSS Index (values: iq, ossindex, enter for exit)? iq
What is the address of your Nexus IQ Server (default: http://localhost:8070)? 
What username do you want to authenticate as (default: admin)? 
What token do you want to use (default: admin123)? 

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!! WARNING : You are using the default username and password for Nexus IQ. !!!!
!!!! You are strongly encouraged to change these, and use a token.           !!!!
!!!! Would you like to change them and try again?                            !!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

[y/N]? y
What is the address of your Nexus IQ Server (default: http://localhost:8070)? 
What username do you want to authenticate as (default: admin)? 
What token do you want to use (default: admin123)? 

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!! WARNING : You are using the default username and password for Nexus IQ. !!!!
!!!! You are strongly encouraged to change these, and use a token.           !!!!
!!!! Would you like to change them and try again?                            !!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

[y/N]? N
```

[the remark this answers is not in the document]
zendern: I like the flowand overall it's good. Might tweak this 
`You are using the default username and password `

From an **and** to an **and/or** since you could change one and not the other.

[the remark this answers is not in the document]
DarthHater: Did that for you and changed the check to a `||` as well

DarthHater:
@zendern let er rip. I had NEVER tested something that took `stdin` input before, fun learning experiment doing so. Tests pass, and confirm that something is written to disk. I can think of one nice thing to do which is to not use `/tmp` like I am and use `ioutils.TempFile`, figured I'd see if you had any other ideas there though!

DarthHater:
I believe I covered all the bases now. Parse tests now set `ConfigLocation` to a nonsensical place so they won't test loading config from file. Previously if you'd set your config locally you'd get failing tests from it loading. 

As well, I changed the logger to truncate the file by default now, which is effectively poor person log rotation. This was not specific to this task, but was easy and I figured I'd do it to make my own testing a bit easier.

I've tested the config for OSS Index and IQ Server myself locally, and everything checks out!

configuration/parse.go
```diff
@@ -91,6 +105,32 @@ Options:
 	return config, nil
 }
 
+func loadConfigFromFile(configLocation string, config *Configuration) error {
+	b, err := ioutil.ReadFile(configLocation)
```
bhamail, line 109: Another thought (I'll add it to #111) is to refactor the `loadConfigFromFile()` and `loadIQConfigFromFile()` methods to take `io.Reader` - instead of a path `configLocation`. The tests could then easily pass in bytes representing the various file contents. This would also allow the test avoid pesky disk access.

configuration/set.go
```diff
@@ -0,0 +1,171 @@
...
+	return
+}
+
+func marshallAndWriteToDisk(config interface{}) (err error) {
+	d, err := yaml.Marshal(config)
+	if err != nil {
+		return err
+	}
+
+	base := filepath.Dir(ConfigLocation)
```
bhamail, line 147: Added @todo for Issue #111 : use `io.Writer` instead of `ConfigLocation` inside `configuration/set.go -> marshallAndWriteToDisk`.

I have Readers/Writers on the brain. ;)

configuration/set.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return
+	}
+
+	LogLady.WithField("config_location", ConfigLocation).Info("Successfully wrote config to disk")
+	fmt.Println(fmt.Sprintf("Successfully wrote config to: %s", ConfigLocation))
+	return
+}
+
+func warnUserOfBadLifeChoices() {
```
bhamail, line 163: :) Oh that it were automatable and resurseable on oneself. Performance hot spot?

configuration/set_test.go
```diff
@@ -0,0 +1,61 @@
+package configuration
```
zendern, line 1: Is it worth adding a test to validate we do the default username/password check and reprompts??

brent-spiner: Added in: 10a38d3 AND fixed an issue with it, great suggestion Lord Zender!

brent-spiner: Added in: 10a38d3 AND fixed an issue with it, great suggestion Lord Zender!

DarthHater: Added in: 10a38d3 AND fixed an issue with it, great suggestion Lord Zender!

PR #112, opened 2020-04-10
docs: apply copyright header

nblair:
This pull request uses our internal tooling (based off of the license-maven-plugin) to consistently apply the copyright at the top of source in this repository.

cc @bhamail / @DarthHater


zendern:
Or even just work it into ci someway ??

DarthHater:
@zendern we have it setup on the "private" side, I imagine we could come up with something for it that's public though. The private side uses a maven plugin to accomplish it, which is funny at some level

zendern:
> @zendern we have it setup on the "private" side, I imagine we could come up with something for it that's public though. The private side uses a maven plugin to accomplish it, which is funny at some level

Cool... yeah even as simple as a bot that checks for licenses at build time would be cool. Speaking of java tooling I’ve used checkstyle to validate license compliance and force manual addition of the headers cause the build would not pass.

Here is an example config
https://github.com/micronaut-projects/micronaut-acme/tree/master/config/checkstyle


And then the gradle magic that wires up that config 

https://github.com/micronaut-projects/micronaut-acme/blob/69f8d0f11ead4b25f731d610efaea49bcd356ec1/build.gradle#L167

fitzoh:
@nblair Are you planning on having jenkins apply checks on this repo?

DarthHater:
@fitzoh we have jenkins setup to do a IQ policy scan, and this too (since it comes for free)

fitzoh:
Cool, just making sure there weren't plans to add a github check that external people can't see

DarthHater:
@fitzoh negative, that's not my plan, and to be honest why I avoided doing a lot of this work. I love the public nature of this project and adding anything that others can't see or benefit from doesn't excite me.

nblair:
> is there a command that was used to automate this update? Could we document said command?

There is, but it depends on a Sonatype docker instance. Let me see if I can cook up a public equivalent.

nblair:
The command uses the [license-maven-plugin](http://mycila.mathieu.photography/license-maven-plugin/). What do you think of adding a thin pom.xml with the required configuration and the maven wrapper to execute?

Another option would be to craft some sort of bot to automatically apply header changes and commit them to master.

Other ideas?

fitzoh:
Might be worth looking at https://github.com/google/addlicense

nblair:
> Might be worth looking at https://github.com/google/addlicense

Nice find @fitzoh - that does the trick, see ca5507b (cc @bhamail).

README.md
```diff
@@ -439,6 +456,10 @@ $ curl -o /path/where/you/want/nancy \
 
 Tests can be run like `go test ./... -v`
 
+Adding new files? Get the license header correct with:
+
+> addlicense -f ./header.txt .
```
zendern, line 461: So this didn't just work for me out of the box..... either we should add to the docs that you have to install `addlicense` (go get -u github.com/google/addlicense) or update the CI build to just do it for you :)

nblair: I'm a go newb :) I did run that `go get` command, and saw that it updated `go.sum` and `go.mod` and assumed it automatically replicated for other development environments. Wasn't sure how those relate to typical build - will each person that clones the repo need to call that `go get` command for that library specifically?

DarthHater: Yes, but we could include it in our `Makefile` more than likely, and just have it run with each build.

zendern: yeah so in this case since what you want is a binary so you can run it.... `go get` will do 2 things..... 1 install the binary into ~/go/bin and add the libs to your project since we are using `go mod` so that you can use them programmatically. If you run a `go mod tidy` which we do on at CI time.....it'll remove addlicense since its not used by the code. So for us to use it locally ever dev will have to install it manually to run that `addlicense` command above. Thus the CI comment..... more stuff there means no one has to worry about it and it just happens :) But just docs are good too for now. Your call

nblair: Thanks @zendern - I ran `go mod tidy` and documented the `go get` command, I'll leave the CI changes to @DarthHater and @bhamail ;)

zendern: @nblair awesome!!! Would you mind writing up an issue to wire this `addlicense` stuff up into CI for us??

nblair: 👍 done in #128 :)

PR #114, opened 2020-04-11
Let's make config fun

### LINKED ISSUE #95: Simplify Config

Right now the methods for parsing config (OSS Index and IQ) are largely identical, which is a code smell!

We should endeavor to make these as simple as possible (but no simpler, don't go too wild!)

Long term code maintainability, right now if we add a common config variable we have two places to deal with it. An example is the verbose flags for the logger.

Buyer's choice! Don't let me hamper your creativity!

* **Anything else?**

AH THE JOYS OF CONFIG!

DarthHater:
Basically, Config was becoming a pretty messy package. This PR aims to make it less messy (in the long term), maintains API compatibility for anyone consuming upstream, and makes another couple improvements while it's at it

* Switches to using a `Config` struct, and making the `Parse` a method off of it
* Since `Config` is largely generic, creates new functions and methods that accept this more generic struct
* Adds new functions that accepts `packageurl.PackageURL` rather than a slice of strings for purls
* Switches everything in `main.go` to use the latest and greatest

Still left to do:
- As always, tests.

* Fixes #95 

cc @bhamail / @DarthHater / @zendern


main.go
```diff
@@ -42,15 +43,18 @@ import (
 func main() {
 	LogLady.Info("Starting Nancy")
 
+	config := configuration.Config{}
+
 	if len(os.Args) > 1 && os.Args[1] == "iq" {
 		LogLady.Info("Nancy parsing config for IQ")
-		config, err := configuration.ParseIQ(os.Args[2:])
+		config.Type = "iq"
```
zendern, line 50: This `setting of type` on the shared config object feels weird to me. Maybe make a "constructor" (yes i know golang doesn't have those) but basically to create a config you must pass a type of config. 

Something like 
```
func NewConfig(string typeOfConfig) *Config{
    config := configuration.Config{}
    //maybe some validation that "iq" or "ossindex" was passed??
    config.Type = typeOfConfig
    return config
}
```

configuration/old_parse.go
```diff
@@ -0,0 +1,304 @@
...
+		case "text":
+			setTextOutputFormat()
+		case "csv":
+			OssIndexOutputFormats[outputFormat].(*audit.CsvFormatter).Quiet = &config.Quiet
+		default:
+			setTextOutputFormat()
+		}
+		config.Formatter = OssIndexOutputFormats[outputFormat]
+	} else {
+		width, _, err := terminal.GetSize(int(os.Stdout.Fd()))
```
zendern, line 232: Didn't know that was a thing.....also would have never looked in a crypto package `golang.org/x/crypto/ssh/terminal`!!?!?? Real odd but neat ¯\_(ツ)_/¯

configuration/parse.go
```diff
@@ -260,14 +236,44 @@ func determineIfLineIsExclusion(ogLine string, config *Configuration) error {
...
+
+	if OssIndexOutputFormats[outputFormat] != nil {
+		switch outputFormat {
+		case "text":
+			setTextOutputFormat()
+		case "csv":
+			OssIndexOutputFormats[outputFormat].(*audit.CsvFormatter).Quiet = &c.Quiet
+		default:
+			setTextOutputFormat()
 		}
```
zendern, line 265: I dont understand the default case here?? If it not text...we set the text output formats?? So the only cases we would hit the default would be "json" and "json-pretty"??

configuration/old_parse.go
```diff
@@ -0,0 +1,304 @@
+// Copyright 2020 Sonatype Inc.
```
zendern, line 1: So this will die at 1.0 time is the thought?? I'd write up another card or tack into the existing....clean up deprecations before 1.0.

iq/old_iq.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020 Sonatype Inc.
```
zendern, line 1: So this will die at 1.0 time is the thought?? I'd write up another card or tack into the existing....clean up deprecations before 1.0.

ossindex/old_ossindex.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2018 Sonatype Inc.
```
zendern, line 1: So this will die at 1.0 time is the thought?? I'd write up another card or tack into the existing....clean up deprecations before 1.0.

configuration/parse.go
```diff
@@ -30,201 +31,176 @@ import (
...
+	UseStdIn    bool
+	NoColor     bool
+	Quiet       bool
+	CleanCache  bool
+	CveList     types.CveListFlag
+	Path        string
+	Formatter   logrus.Formatter
+	Username    string `yaml:"Username"`
+	Token       string `yaml:"Token"`
 }
```
zendern, line 59: So I like the idea of the singular config object but not a fan of all properties jammed in there without knowing what is what. What if we use composition here and at least make a clearer distinction on what properties are for what and what is shared. 

```
type Config struct {
	Help       bool		
        Type        string
	NoColor    bool	
	Quiet      bool	
	Version    bool	
	CleanCache bool	
	CveList    types.CveListFlag	
	Path       string	
	Info       bool	
	Debug      bool	
	Trace      bool	
	Username   string `yaml:"Username"`	
	Token      string `yaml:"Token"`	
        IqConfiguration
        OssIndexConfiguration
}	

type IqConfiguration struct {	
	Stage       string
	Application string
	Server      string
	MaxRetries  int
}

type OssIndexConfiguration struct {	
	Formatter  logrus.Formatter	
}
```

packages/dep.go
```diff
@@ -18,6 +18,7 @@ import (
 
 	"github.com/Masterminds/semver"
 	"github.com/golang/dep"
+	"github.com/package-url/packageurl-go"
```
zendern, line 21: 👍

main.go
DarthHater: That's a great suggestion. I sorta just jammed this in here to get it working. I'll noodle on that!

configuration/old_parse.go
DarthHater: It was driving me nuts, haha. The OCD in my brain was like BUT THIS IS NOT FULL TERMINAL WIDTH.

configuration/old_parse.go
DarthHater: Yep! That's the idea. I'll create one.

configuration/parse.go
DarthHater: That's great! I may actually split the username and token into those structs too, because they are fully distinct.

DarthHater: I think I got part of this in: 37493b2

main.go
DarthHater: I think I got this in: 37493b2

Went a lil further, it only accepts an "enum" of sorts, in my next pass I can add some logic that returns an error if you send in the wrong one.

DarthHater:
Newest changes let you have IQ and OSS Index config, loads them both, and worked for me when calling IQ Server, which is great :)

configuration/parse.go
DarthHater: I'll move the other fields in, just gotta figure out if `go-yaml` will let me use `omitempty` to keep them from being persisted when we write to the file system (none of them are really global config, more like single use config)

configuration/old_parse.go
DarthHater: https://github.com/sonatype-nexus-community/nancy/issues/126

iq/old_iq.go
DarthHater: https://github.com/sonatype-nexus-community/nancy/issues/126

ossindex/old_ossindex.go
DarthHater: https://github.com/sonatype-nexus-community/nancy/issues/126

configuration/parse.go
DarthHater: I was flying way to close to the sun, Icarus style: bee0c6f

configuration/parse.go
DarthHater: I think I finished up with moving stuff, here: 07117d3 (CleanCache could be thought of as just OSS Index specific, but I'm on the fence)

bhamail: So I'm still wrapping my head around all this, so ignore if this is crazy talk, but this reminded me of config code from the nfpm project (guts of goreleaser). They used a sort of "base" struct (named [Info](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L108)) to hold common config, and then used an [Overridables](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L129) struct for sub-type specific stuff. Could go deeper with type specific structs for [Rpm](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L145), [Deb](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L151). The secret sauce is how these get merged together via: [mergo](https://github.com/imdario/mergo). See how nfpm uses the lib [here](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L83) and [also here](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L91) 

(I wad [bitten](https://github.com/goreleaser/nfpm/pull/126#issuecomment-611721607) when I did the structs wrong, but it does seem pretty powerful)

bhamail:
Do we still need these changes after converting to Cobra/Viper? Or is it all addressed in #161 ?

DarthHater:
Naw pretty sure this can go away @bhamail !

PR #129, opened 2020-04-13
License check

### LINKED ISSUE #128: Add license check to CircleCI configuration

In #112, we added configuration and documentation on how to add the correct license header on all source files. We'd like the CircleCI builds to automate that check.

Be good open source citizens and keep our license documented correctly.

Not sure!

* **Anything else?**

cc @bhamail / @DarthHater

DarthHater:
For my good pal @nblair , here's `addlicense` in CircleCI via our `Makefile` to please Lord Zender.

Example build with this applied:
https://circleci.com/gh/sonatype-nexus-community/nancy/237

* Modifies `Makefile` to have a `headers` command
* Installs `github.com/google/addlicense` in that command
* Runs `addlicense` like so: `addlicense -check -f ./header.txt .`

Fails on this branch because a few files ARE missing headers. However, a few of them I have my doubts on needing one:

```
Dockerfile ? Does this really need a license header?
.goreleaser.yml ? same query
.circleci/config.yml ? same query
```
As well, addlicense seems to miss `Dockerfile.goreleaser` because it doesn't know about that extension.

* Fixes #128 

cc @bhamail / @DarthHater

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,7 @@ jobs:
           keys:
             - go-mod-v1-{{ checksum "go.sum" }}
       - run: go get -u github.com/jstemmer/go-junit-report
+      - run: make headers
       - run: make deps
       - run: make ci-lint
       - run: make test | go-junit-report > $TEST_RESULTS/gotest/report.xml
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,17 @@
+# Copyright 2018-present Sonatype Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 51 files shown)


zendern:
hmmmm....yeah im not the license police but putting that in the build files feels odd. But its also probably not the end of the world. ¯\_(ツ)_/¯ 

re: `Dockerfile.goreleaser` Looks like this is the magic it wants... maybe we change the name to appease it??
https://github.com/google/addlicense/blob/7c013a14f2e2f10021b6a6876fc54295f0c4ad82/main.go#L214-L215

DarthHater:
@zendern this is now good to go, @nblair too. Approve whenever you want!

PR #127, opened 2020-04-13
Initial cleanup of error messages

designergod420:
Working with @DarthHater we went through Nancy and looked for any instances where we were using fmt to print an error and either replaced with a customerror.check or logged with LogLady. 

It relates to the following issue #s:
* Fixes #115 

cc @bhamail / @DarthHater

PR #130, opened 2020-04-14
Give all version info

### LINKED ISSUE #125: Give all version info and more to the LogLady
@fitzoh reported issue #124 and even though we have more logging and thats good .... its not super useful to only know that the version was 0.1.17. We could log more details about the actual binary version (all the version info we have). Other details such as os version, cpu architecture,   even what binary cpu architecture nancy is currently being used (32 bit/64 bit version). 

Better debugging capabilities

LogLady will lead your way  

* **Anything else?**

cc @bhamail / @DarthHater

protoworlock69:
Populated architecture and operating system based on run time package 

* Fixes #125

cc @bhamail / @DarthHater

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


DarthHater:
@zendern @bhamail you think we need any more bits than this? This seemed like enough to me to deduce what is going on. MAYBE OS version?

zendern:
@protoworlock69 @DarthHater Changes look good overall. Yeah if you can get OS version sure why not.... but where those details where added doesn't help a ton now that we trunc the file every time. Flow right now would be if you ask the user to dump the logs they will have to explicitly capture them once and then run version command and capture that set. Was hoping for it all in one go. 

https://github.com/sonatype-nexus-community/nancy/blob/209a15f9d9bb83cf9a42cd9809d81b304b05a797/main.go#L80
^^ I think this is where we would also want to update.

DarthHater:
Sounds good!

bhamail:
While looking into this, I noticed an odd message in the beginning of the output:

```
open /Users/bhamail/.ossindex/.oss-index-config: no such file or directory
```

I'll look into why that path is being executed, maybe create an Issue and/or PR.

DarthHater:
@bhamail check #127 , that PR should at least drop the fmt that causes that!

protoworlock69:
@zendern feedback addressed, Jeffry and I looked at getting Operating System version and decided it was more trouble than it was worth (varies due to cross platform issues).

PR #54, opened 2019-11-12
Badgers we dont need no stinking badgers

DarthHater:
`go get` was upgrading badger, and everything was sad, now it's not! Badger upgraded.

This pull request makes the following changes:
* Switches to calling DefaultOptions as a function with the db path
* Switches to using an Entry struct and then setting TTL on this as SetWithTTL was removed from Badger 1.6 
* Updates test

Could use some better testing than the quick "Yeah it works" that I gave it! 

It relates to the following issue #s:
* Fixes #18 (sorta, the current case someone asked about)

cc @bhamail / @DarthHater / @allenhsieh

Issue #79, opened 2020-02-22
Badger alternative


zendern:
* **What are you trying to do?**
Find a caching alternative for our current Badger usage. As part of #77 Badger issues/corruption was brought up and the fact that its probably more heavyweight than what we really need. On top of that we have other issues currently with later versions of Badger having vulnerabilities itself #54. So removing it would hopefully resolve those issues. 

By removing Badger it should hopefully fix this as well #64  .

* **What feature or behavior is this required for?**
Caching of the results from OSS Index. It is rate limited so the reason this is in place is to not overload that service and realistically deps don't change that often so caching the results for 12 hours currently helps with those things. 

* **How could we solve this issue? (Not knowing is okay!)**
Does go have a simple http request diskcache??
Serialize the contents to disk ourselves??
Find a different caching solution instead of Badger??

* **Anything else?**

cc @bhamail / @DarthHater

fitzoh:
>Does go have a simple http request diskcache??

* https://github.com/victorspringer/http-cache
no disk based adapter (although badger is on the roadmap 😆 )
* https://github.com/gregjones/httpcache
This one has a couple disk based backends:
  * github.com/gregjones/httpcache/memcache provides memcache implementations, for both App Engine and 'normal' memcache servers.
  * github.com/gregjones/httpcache/leveldbcache provides a filesystem-backed cache using leveldb.
  * github.com/birkelund/boltdbcache provides a BoltDB implementation (based on the bbolt fork).

>Serialize the contents to disk ourselves??

This feels easy to mess up

>Find a different caching solution instead of Badger??

* https://github.com/etcd-io/bbolt is the promariy maintained fork of https://github.com/boltdb/bolt
* sqlite might be worth considering
  * if corruption is a major concern, this is probably the most battle tested option out there
  * `DELETE FROM <cache_table>` makes cache cleanup easy
  * performance is relatively unimportant here

>Anything else?

Maybe @DarthHater can hook us up w/ a higher rate limit if we use a Nancy  user agent and forget about caching altogether.

DarthHater:
Rate limiting is more or less here to stay, doubt we'd go that route @fitzoh (although I love the suggestion).

We should add "works on WSL (windows subsystem for linux)", etc... to our requirements, as Badger doesn't seem to work well in that context

zendern:
#64 ^^^ As @DarthHater said above .... hopefully after this is played this issue will be resolved as well.

DarthHater:
[...]

PR #133, opened 2020-04-15
RIP Badger

DarthHater:
The time has come, let's simplify our caching DB

* Moves cache functionality into `/ossindex/internal/cache`
* Adds two protected functions only accessible to us, `func Insert(coordinates []types.Coordinate) (err error)` and `func GetCacheValues(purls []string) ([]string, []types.Coordinate, error)` that accomplish inserting a value into the cache, and then getting (and removing if it doesn't exist)
* Rips out Badger and replaces it with `https://github.com/recoilme/pudge`, which is written entirely with the stdlib, and looks pretty good in use so far
* Sets tests back to run in parallel, since we no longer have a lockfile issue
* Allows for TTL to be configured so that a) easier to test and b) we can allow users to set it later
* Allows for DBName to be configured to that a) easier to test

* Fixes #54 #79 #3 

cc @bhamail / @DarthHater


ossindex/internal/cache/cache.go
```diff
@@ -0,0 +1,168 @@
...
+	. "github.com/sonatype-nexus-community/nancy/logger"
+	"github.com/sonatype-nexus-community/nancy/types"
+)
+
+// DBName is exported from cache so that in tests we can switch the DB name, and avoid polluting our real DB
+var DBName = "nancy-cache"
+
+// TTL is exported so that you can a) set the TTL to a lower period of time if you want, and b)
+// so that in tests we can simulate an expired cache object
+var TTL = time.Now().Local().Add(time.Hour * 12)
```
fitzoh, line 40: I would toss these into a struct and convert all the functions into methods of that struct.

```go
type Cache struct{
    DbName string
    TTL time.Duration
}

//instead of
func InsertValuesIntoCache(coordinates []types.Coordinate, ttl time.Time) (err error){

}

//it would be something like
func (c *Cache) Insert(coordinates []types.Coordinate, ttl time.Time) (err error){

}

```

DarthHater: In: 45bf59a and 1d02418

DarthHater: Something I don't like: you have to explicitly set the TTL and DBName, and I hate doing `New()` functions, but it's likely ok since this package is internal and for our use only.

ossindex/internal/cache/cache.go
```diff
@@ -0,0 +1,172 @@
...
+	doSet := func(coordinate types.Coordinate) error {
+		err = pudge.Set(c.getDatabaseDirectory(), strings.ToLower(coordinate.Coordinates), DBValue{Coordinates: coordinate, TTL: c.TTL.Unix()})
+		if err != nil {
+			LogLady.WithField("error", err).Error("Unable to add coordinate to cache DB")
+			return err
+		}
+		return nil
+	}
+
+	for i := 0; i < len(coordinates); i++ {
```
fitzoh, line 99: can you switch to a range here?

DarthHater: Negative, can't use `continue` in a range, for some very peculiar reason (I had it set as one earlier and I was like WHY IS THIS NOT BREAKING!)

ossindex/internal/cache/cache.go
```diff
@@ -0,0 +1,172 @@
...
+type DBValue struct {
+	Coordinates types.Coordinate
+	TTL         int64
+}
+
+func (c *Cache) getDatabaseDirectory() (dbDir string) {
+	usr, err := user.Current()
+	customerrors.Check(err, "Error getting user home")
+
+	return path.Join(usr.HomeDir, types.OssIndexDirName, dbDirName, c.DBName)
```
fitzoh, line 54: naming feels a little off here.
I would think that `dbDirName` would be the last part of a path of `getDatabaseDirectory`.

Also does the directory make sense as part of the `Cache` struct?

ossindex/internal/cache/cache.go
fitzoh: ha, better you than me :D

ossindex/internal/cache/cache.go
DarthHater: Adjusted a few things: 760310f

ossindex/internal/cache/cache_test.go
```diff
@@ -0,0 +1,122 @@
...
+	newPurls, results, err := cache.GetCacheValues(purls)
+
+	assert.Equal(t, purls, newPurls)
+	assert.Empty(t, results)
+	assert.Nil(t, err)
+
+	var result DBValue
+	err = cache.getKeyAndHydrate(purls[0], &result)
+	assert.NotNil(t, err)
+	if _, ok := err.(*os.PathError); ok {
```
bhamail, line 80: minor: Not sure I understand why we check for "not" os.PathError here. Is the intent to check for some other specific error type? I would expect to verify a "key not found" or some such error.

ossindex/internal/cache/cache_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/shopspring/decimal"
+	"github.com/sonatype-nexus-community/nancy/types"
+	"github.com/stretchr/testify/assert"
+)
+
+var coordinates []types.Coordinate
+
+var purls []string
+
```
bhamail, line 33: A big Thank You for these tests. It really helps me get up to speed doing review.

ossindex/ossindex_test.go
```diff
@@ -184,74 +137,82 @@ func verifyClientCallAndWriteValidPackageResponse(t *testing.T, r *http.Request,
...
+				Description: "description",
+				CvssScore:   dec,
+				CvssVector:  "vectorvictor",
+				Cve:         "CVE-123-123",
+				Reference:   "http://www.internet.com",
+				Excluded:    false,
+			},
+		},
+	}
+	dbCache.DBName = "nancy-test"
```
bhamail, line 209: Noice!

ossindex/internal/cache/cache_test.go
```diff
@@ -0,0 +1,122 @@
...
+				Reference:   "http://www.internet.com",
+				Excluded:    false,
+			},
+		},
+	}
+
+	purls = append(purls, "pkg:golang/test@0.0.0")
+
+	coordinates = append(coordinates, coordinate)
+	cache := Cache{DBName: "nancy-cache-test", TTL: time.Now().Local().Add(time.Hour * 12)}
```
bhamail, line 109: Huzzah! Luvin' the separate DBName for tests.

ossindex/internal/cache/cache_test.go
DarthHater: This check looks for an error (see assert.NotNil() above), and verifies that it's type of `*os.PathError` as that is what is sent back when a key is not found. After `GetCacheValues` is run, if an expired key is run into, it deletes it from Pudge, so it verifies that happens inside the method (a lil granular, but I figured it was worth it)

DarthHater: And turns out YOU RIGHT: e9f2708

ossindex/internal/cache/cache_test.go
DarthHater: Yeah, I figured it could help for them all running in parallel.

Issue #78, opened 2020-02-22
Make nancy not so exit heavy


zendern:
* **What are you trying to do?**
As part of #68 @DarthHater was trying to do some testing and it the fact that Nancy exits like its going out of style made it overly difficult. 

For now the exits have been put back in place instead of using a panic which made testing easier to do. 

General go guidance is as follows, so this is where we should start to move the codebase. 
```
The convention in the Go libraries is that even when a package uses panic internally, its external API still presents explicit error return values.
```

* **What feature or behavior is this required for?**
No real feature just more of a tech debt/cleanup/make it better type thing.

* **How could we solve this issue? (Not knowing is okay!)**
Stop exiting so much and panic instead but at the same time we still will want to preserve the exit codes and the user friendly messages after this change. No stack trace should be presented to the user. 

* **Anything else?**
¯\_(ツ)_/¯ 

cc @bhamail / @DarthHater

DarthHater:
@bhamail good issue to tack on to for #134

bhamail:
@DarthHater Agreed. Will do

bhamail:
Working on this here: https://github.com/sonatype-nexus-community/nancy/pull/137

PR #134, opened 2020-04-15
WIP, POC Cobra. Am I crazy?

bhamail:
I expect I'm missing something, but if you pull down this branch, run:

```
go build
go list -m all | ./nancy --quiet
open /Users/bhamail/.ossindex/.oss-index-config: no such file or directory
------------------------------------------------------------
[107/129]pkg:golang/golang.org/x/crypto@0.0.0-20190308221718-c2843e01d9a2  [Vulnerable]   1 known vulnerabilities affecting installed version

[CVE-2019-11840]  Use of Insufficiently Random Values
An issue was discovered in supplementary Go cryptography libraries, aka golang-googlecode-go-crypto, before 2019-03-20. A flaw was found in the amd64 implementation of golang.org/x/crypto/salsa20 and golang.org/x/crypto/salsa20/salsa. If more than 256 GiB of keystream is generated, or if the counter otherwise grows greater than 32 bits, the amd64 implementation will first generate incorrect output, and then cycle back to previously generated keystream. Repeated keystream bytes can lead to loss of confidentiality in encryption applications, or to predictability in CSPRNG applications.

ID:5121f5ff-9831-44a6-af2e-24f7301d1df7
Details:https://ossindex.sonatype.org/vuln/5121f5ff-9831-44a6-af2e-24f7301d1df7
Audited dependencies:129,Vulnerable:1
```
It works.

Also, loud works too:
```
go list -m all | ./nancy
```

Seems like stdin is working with Cobra now?!?

Only implemented OSSI (haven't tried IQ yet, but wouldn't expect a difference).

As a side effect, I will try to address heavy usage of os.Exit() mentioned in #78

cc @bhamail / @DarthHater


DarthHater:
SALUTE THIS DAN

bhamail:
So it appears nobody strongly objects to switching to Cobra, so I'm gonna go for it.

I started down the path, and quickly hit a wall due to the scattered `os.Exit()` calls making it very hard to test and refactor. I will push up some of the hoops I jumped through, if only for reference here. e.g. Tried a technique (suggested by one of the go devs) that executes the code calling os.Exit() in a separate process. Wild and crazy, but works. Another side effect of my attempts to tame os.Exit() produced a method to get the name of the calling function, as well as an `exiter` interface that can mock the exit when needed.

My plan is to first refactor the current code to remove some of the `os.Exit()` and replace with call stacks that return `error`. I'll create a separate PR for that.

bhamail:
@zendern 
JUST A NOTE OF INTENT (BTW, IT'S CAPS LOCK FRIDAY):

WHILE CONVERTING TO USE COBRA COMMANDS, I NOTICED OUR
CURRENT SINGLE HYPHEN/MULTI-CHARACTER FLAGS NEED TO CHANGE:
-output=json TO --output=json

JEFFRY IS ALSO WORKING ON SOME REFACTORING OF SHARED CODE INTO A NEW PROJECT.

FOR THESE REASONS, I EXPECT WE WILL BE BUMPING THE VERSION OF NANCY TO 1.0.0. 
HOLLER IF YOU HAVE CONCERNS. (BETTER HOLLER LOUDLY, OR WE WON'T HEAR YOU).

zendern:
@bhamail CAPS LOCK DOWN!! #coronavirus ALSO 👍👍

bhamail:
This work has been merged into branch #161

PR #137, opened 2020-04-18
Exit purge

### LINKED ISSUE #78: Make nancy not so exit heavy

bhamail:
Minimize places where we call os.Exit(), in preparation for conversion to using Cobra #134 

I think this is ready for review.

I was a little concerned about the CI build failing in the logger_test.go. Seemed to be an unrelated change. Rerunning the CI build and it just passed. Boggle. We should keep an eye out in case this is a heisenbug. 

BTW, we may have a decision point regarding "--command, -flag" syntax in the near future.

Fixes #78

cc @bhamail / @DarthHater

### DIFF
--- a/audit/csvformatter.go
+++ b/audit/csvformatter.go
@@ -24,7 +24,6 @@ import (
 	"strconv"
 
 	. "github.com/sirupsen/logrus"
-	"github.com/sonatype-nexus-community/nancy/customerrors"
 	"github.com/sonatype-nexus-community/nancy/types"
 )
 
@@ -57,33 +56,56 @@ func (f *CsvFormatter) Format(entry *Entry) ([]byte, error) {
 		var buf bytes.Buffer
 		w := csv.NewWriter(&buf)
 
-		f.write(w, []string{"Summary"})
-		f.write(w, summaryHeader)
-		f.write(w, summaryRow)
+		var err error
+		if err = f.write(w, []string{"Summary"}); err != nil {
+			return nil, err
+		}
+		if err = f.write(w, summaryHeader); err != nil {
+			return nil, err
+		}
+		if err = f.write(w, summaryRow); err != nil {
+			return nil, err
+		}
 
 		if !*f.Quiet {
 			invalidCount := len(invalidEntries)
 			if invalidCount > 0 {
-				f.write(w, []string{""})
-				f.write(w, []string{"Invalid Package(s)"})
-				f.write(w, invalidHeader)
+				if err = f.write(w, []string{""}); err != nil {
+					return nil, err
+				}
+				if err = f.write(w, []string{"Invalid Package(s)"}); err != nil {
+					return nil, err
+				}
+				if err = f.write(w, invalidHeader); err != nil {
+					return nil, err
+				}
 				for i := 1; i <= invalidCount; i++ {
 					invalidEntry := invalidEntries[i-1]
-					f.write(w, []string{"[" + strconv.Itoa(i) + "/" + strconv.Itoa(invalidCount) + "]", invalidEntry.Coordinates, "Does not use SemVer"})
+					if err = f.write(w, []string{"[" + strconv.Itoa(i) + "/" + strconv.Itoa(invalidCount) + "]", invalidEntry.Coordinates, "Does not use SemVer"}); err != nil {
+						return nil, err
+					}
 				}
 			}
 		}
 
 		if !*f.Quiet || numVulnerable > 0 {
-			f.write(w, []string{""})
-			f.write(w, []string{"Audited Package(s)"})
-			f.write(w, auditedHeader)
+			if err = f.write(w, []string{""}); err != nil {
+				return nil, err
+			}
+			if err = f.write(w, []string{"Audited Package(s)"}); err != nil {
+				return nil, err
+			}
+			if err = f.write(w, auditedHeader); err != nil {
+				return nil, err
+			}
 		}
 		for i := 1; i <= len(auditedEntries); i++ {
 			auditEntry := auditedEntries[i-1]
 			if auditEntry.IsVulnerable() || !*f.Quiet {
 				jsonVulns, _ := json.Marshal(auditEntry.Vulnerabilities)
-				f.write(w, []string{"[" + strconv.Itoa(i) + "/" + strconv.Itoa(packageCount) + "]", auditEntry.Coordinates, strconv.FormatBool(auditEntry.IsVulnerable()), strconv.Itoa(len(auditEntry.Vulnerabilities)), string(jsonVulns)})
+				if err = f.write(w, []string{"[" + strconv.Itoa(i) + "/" + strconv.Itoa(packageCount) + "]", auditEntry.Coordinates, strconv.FormatBool(auditEntry.IsVulnerable()), strconv.Itoa(len(auditEntry.Vulnerabilities)), string(jsonVulns)}); err != nil {
+					return nil, err
+				}
 			}
 		}
 
@@ -96,7 +118,6 @@ func (f *CsvFormatter) Format(entry *Entry) ([]byte, error) {
 
 }
 
-func (f *CsvFormatter) write(w *csv.Writer, line []string) {
-	err := w.Write(line)
-	customerrors.Check(err, "Failed to write data to csv")
+func (f *CsvFormatter) write(w *csv.Writer, line []string) error {
+	return w.Write(line)
 }
--- a/audit/csvformatter.go
+++ b/audit/csvformatter.go
@@ -21,6 +21,7 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"errors"
+	"github.com/sonatype-nexus-community/nancy/customerrors"
 	"strconv"
 
 	. "github.com/sirupsen/logrus"
@@ -119,5 +120,8 @@ func (f *CsvFormatter) Format(entry *Entry) ([]byte, error) {
 }
 
 func (f *CsvFormatter) write(w *csv.Writer, line []string) error {
-	return w.Write(line)
+	if err := w.Write(line); err != nil {
+		return customerrors.NewErrorExitPrintHelp(err, "Failed to write data to csv")
+	}
+	return nil
 }
--- a/configuration/parse.go
+++ b/configuration/parse.go
@@ -70,7 +70,7 @@ var unixComments = regexp.MustCompile(`#.*$`)
 var untilComment = regexp.MustCompile(`(until=)(.*)`)
 
 func ParseIQ(args []string) (config IqConfiguration, err error) {
-	iqCommand := flag.NewFlagSet("iq", flag.ExitOnError)
+	iqCommand := flag.NewFlagSet("iq", flag.ContinueOnError)
 	iqCommand.BoolVar(&config.Info, "v", false, "Set log level to Info")
 	iqCommand.BoolVar(&config.Debug, "vv", false, "Set log level to Debug")
 	iqCommand.BoolVar(&config.Trace, "vvv", false, "Set log level to Trace")
@@ -88,7 +88,8 @@ func ParseIQ(args []string) (config IqConfiguration, err error) {
 Options:
 `)
 		iqCommand.PrintDefaults()
-		os.Exit(2)
+		//os.Exit(2)
+		// Avoid exit calls, but caller should use exit code 2 when handling
 	}
 
 	ConfigLocation = filepath.Join(HomeDir, types.IQServerDirName, types.IQServerConfigFileName)
@@ -103,7 +104,7 @@ Options:
 		return config, err
 	}
 
-	return config, nil
+	return config, err
 }
 
 func loadConfigFromFile(configLocation string, config *Configuration) error {
@@ -173,7 +174,8 @@ func Parse(args []string) (Configuration, error) {
 Options:
 `)
 		flag.PrintDefaults()
-		os.Exit(2)
+		// os.Exit(2)
+		// Avoid exit calls, but caller should use exit code 2 when handling
 	}
 
 	ConfigLocation = filepath.Join(HomeDir, types.OssIndexDirName, types.OssIndexConfigFileName)
--- a/configuration/parse.go
+++ b/configuration/parse.go
@@ -88,8 +88,6 @@ func ParseIQ(args []string) (config IqConfiguration, err error) {
 Options:
 `)
 		iqCommand.PrintDefaults()
-		//os.Exit(2)
-		// Avoid exit calls, but caller should use exit code 2 when handling
 	}
 
 	ConfigLocation = filepath.Join(HomeDir, types.IQServerDirName, types.IQServerConfigFileName)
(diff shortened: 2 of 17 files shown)


DarthHater:
I'm sure that you still are cranking, but when I run `./nancy iq` I get a SIGSEGV: (after usage).

```
$ ./nancy iq
Usage:
        go list -m all | nancy iq [options]

Options:
  -application string
        Specify application ID for request
  -max-retries int
        Specify maximum number of tries to poll Nexus IQ Server (default 300)
  -server-url string
        Specify Nexus IQ Server URL/port (default "http://localhost:8070")
  -stage string
        Specify stage for application (default "develop")
  -token string
        Specify Nexus IQ token/password for request (default "admin123")
  -user string
        Specify Nexus IQ username for request (default "admin")
  -v    Set log level to Info
  -vv
        Set log level to Debug
  -vvv
        Set log level to Trace
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x14fc90f]

goroutine 1 [running]:
github.com/sonatype-nexus-community/nancy/customerrors.ErrorExit.Error(0x0, 0x0, 0x0, 0x0, 0x2, 0xc000213710, 0xc0001a3a18)
        /go/src/github.com/sonatype-nexus-community/nancy/customerrors/error.go:30 +0x3f
github.com/sirupsen/logrus.(*JSONFormatter).Format(0xc000110040, 0xc0002209a0, 0x10703d2, 0x2, 0x0, 0x1056779, 0xc0001a3d78)
        /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/json_formatter.go:67 +0x1cd
github.com/sirupsen/logrus.(*Entry).write(0xc0002209a0)
        /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:277 +0x7c
github.com/sirupsen/logrus.Entry.log(0xc00013e070, 0xc000213710, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:253 +0x1c2
github.com/sirupsen/logrus.(*Entry).Log(0xc000220930, 0x2, 0xc0001a3f58, 0x1, 0x1)
        /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:289 +0xeb
github.com/sirupsen/logrus.(*Entry).Error(...)
        /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:318
main.main()
        /go/src/github.com/sonatype-nexus-community/nancy/main.go:64 +0x20b
```

zendern:
> I'm sure that you still are cranking, but when I run `./nancy iq` I get a SIGSEGV: (after usage).
> 
> ```
> $ ./nancy iq
> Usage:
>         go list -m all | nancy iq [options]
> 
> Options:
>   -application string
>         Specify application ID for request
>   -max-retries int
>         Specify maximum number of tries to poll Nexus IQ Server (default 300)
>   -server-url string
>         Specify Nexus IQ Server URL/port (default "http://localhost:8070")
>   -stage string
>         Specify stage for application (default "develop")
>   -token string
>         Specify Nexus IQ token/password for request (default "admin123")
>   -user string
>         Specify Nexus IQ username for request (default "admin")
>   -v    Set log level to Info
>   -vv
>         Set log level to Debug
>   -vvv
>         Set log level to Trace
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x14fc90f]
> 
> goroutine 1 [running]:
> github.com/sonatype-nexus-community/nancy/customerrors.ErrorExit.Error(0x0, 0x0, 0x0, 0x0, 0x2, 0xc000213710, 0xc0001a3a18)
>         /go/src/github.com/sonatype-nexus-community/nancy/customerrors/error.go:30 +0x3f
> github.com/sirupsen/logrus.(*JSONFormatter).Format(0xc000110040, 0xc0002209a0, 0x10703d2, 0x2, 0x0, 0x1056779, 0xc0001a3d78)
>         /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/json_formatter.go:67 +0x1cd
> github.com/sirupsen/logrus.(*Entry).write(0xc0002209a0)
>         /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:277 +0x7c
> github.com/sirupsen/logrus.Entry.log(0xc00013e070, 0xc000213710, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
>         /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:253 +0x1c2
> github.com/sirupsen/logrus.(*Entry).Log(0xc000220930, 0x2, 0xc0001a3f58, 0x1, 0x1)
>         /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:289 +0xeb
> github.com/sirupsen/logrus.(*Entry).Error(...)
>         /go/pkg/mod/github.com/sirupsen/logrus@v1.5.0/entry.go:318
> main.main()
>         /go/src/github.com/sonatype-nexus-community/nancy/main.go:64 +0x20b
> ```

re: this `Hopefully Jeffry will NOT look at this over the weekend. 😄`
@bhamail I believe the answer is yeah not going to happen 😝

bhamail:
> re: this `Hopefully Jeffry will NOT look at this over the weekend. 😄`
> @bhamail I believe the answer is yeah not going to happen 😝

The was a snowballs chance.... ;)

Fix issue with `nil` Err object causing seg fault: 24f46c0

customerrors/error.go
```diff
@@ -18,33 +18,20 @@ package customerrors
...
+	ExitCode int
 }
 
-func Check(err error, message string) {
-	if err != nil {
-		location, _ := LogFileLocation()
-		myErr := SwError{Message: message, Err: err}
-		LogLady.WithField("error", err).Error(message)
-		fmt.Println(myErr.Error())
-		fmt.Printf("For more information, check the log file at %s\n", location)
```
DarthHater, line 46: I'm not sure exactly where you may have moved this, but this was sorta important. Do you mind linking me to where it's been implemented since this PR is kinda big?

cyclonedx/cyclonedx_test.go
```diff
@@ -24,11 +24,11 @@ import (
 	"github.com/package-url/packageurl-go"
 	"github.com/shopspring/decimal"
 	"github.com/sonatype-nexus-community/nancy/types"
-	assert "gopkg.in/go-playground/assert.v1"
+	"gopkg.in/go-playground/assert.v1"
 )
 
 func TestCreateSBOMFromPackageURLs(t *testing.T) {
-	results := []packageurl.PackageURL{}
+	var results []packageurl.PackageURL
```
DarthHater, line 31: Are these linter changes?

customerrors/error.go
bhamail: It's been replaced by returning `error` and using the customer `ExitError` where a specific exit code needed to bubble up the stack.

cyclonedx/cyclonedx_test.go
bhamail: Yes, I was just going through linter suggestions. Happy to revert if desired.

main.go
```diff
@@ -46,34 +47,65 @@ import (
...
-		LogLady.Info("Nancy parsing config for IQ")
-		config, err := configuration.ParseIQ(os.Args[2:])
-		if err != nil {
-			flag.Usage()
-			os.Exit(1)
-		}
-		LogLady.WithField("config", config).Info("Obtained IQ config")
-		processIQConfig(config)
-		LogLady.Info("Nancy finished parsing config for IQ")
+		err = doIq(os.Args[2:])
```
DarthHater, line 52: I like that you simplified the func name, but now it's a lil impossible to tell what it actually does. `doIq` is a bit too simple, I think. As well, it should likely be `doIQ`, as general golang convention is to capitalize stuff like that (`URL` vs `Url`, etc...)

cyclonedx/cyclonedx_test.go
DarthHater: Naw no big deal, just kinda curious why it happened.

customerrors/error.go
DarthHater: Right, but where's the part that tells someone to look at their log file for more information?

audit/csvformatter.go
```diff
@@ -57,33 +56,56 @@ func (f *CsvFormatter) Format(entry *Entry) ([]byte, error) {
...
+		var err error
+		if err = f.write(w, []string{"Summary"}); err != nil {
+			return nil, err
+		}
+		if err = f.write(w, summaryHeader); err != nil {
+			return nil, err
+		}
+		if err = f.write(w, summaryRow); err != nil {
+			return nil, err
+		}
```
DarthHater, line 68: Holy code duplication batperson. I'm counting 11 or so blocks that are looking almost the same. Any way to simplify this down?

configuration/parse.go
```diff
@@ -88,7 +88,8 @@ func ParseIQ(args []string) (config IqConfiguration, err error) {
 Options:
 `)
 		iqCommand.PrintDefaults()
-		os.Exit(2)
+		//os.Exit(2)
+		// Avoid exit calls, but caller should use exit code 2 when handling
```
DarthHater, line 92: I would remove these comments

configuration/parse.go
```diff
@@ -173,7 +174,8 @@ func Parse(args []string) (Configuration, error) {
 Options:
 `)
 		flag.PrintDefaults()
-		os.Exit(2)
+		// os.Exit(2)
+		// Avoid exit calls, but caller should use exit code 2 when handling
```
DarthHater, line 178: I would remove these comments, too

customerrors/error.go
```diff
@@ -18,33 +18,20 @@ package customerrors
...
-		fmt.Println("nancy version:", buildversion.BuildVersion)
-		myErr.Exit()
+func (ee ErrorExit) Error() string {
+	var errString string
+	if ee.Err != nil {
+		errString = ee.Err.Error()
+	} else {
+		errString = "nil"
```
DarthHater, line 34: Why `"nil"`? There is no such thing as a nil string, why not `""`?

main_test.go
```diff
@@ -33,3 +40,194 @@ func TestBadArgs(t *testing.T) {
...
+	purls := []string{"pkg:github/BurntSushi/toml@0.3.1"}
+	invalidPurls := []string{"invalidPurl"}
+	noColor := true
+	quiet := true
+	config := configuration.Configuration{Formatter: &audit.AuditLogTextFormatter{Quiet: &quiet, NoColor: &noColor}}
+	err := checkOSSIndex(purls, invalidPurls, config)
+	assert.NoError(t, err)
+}
+
+func TestDoOssi(t *testing.T) {
```
DarthHater, line 225: I am always SUPER wary of testing private methods. Do you think we really need to do this, versus an Integration test via dogfooding, etc... that tests more "public" expected behavior?

customerrors/error.go
bhamail: > Right, but where's the part that tells someone to look at their log file for more information?

Ah, I see your point now. I went back through all the places that used to call the old "check" method, and changed them to call a new "constructor" method the prints the old message about where to look in the logs for more info, and returns a new ErrorExit to bubble up the call stack.  
ccc94e1

main.go
bhamail: These `do*` methods will likely map directly to new Cobra `command` structures. I will try to come up with better names for the Cobra commands. For now, the main goal was to simplify the main method enough so I could grok replacing the `os.Exit` stuff with centralized "exit processing" in `main()`.

audit/csvformatter.go
bhamail: Agreed. I'm open to suggestions here. Smells funny, but I'm not seeing a less stanky approach yet.

configuration/parse.go
bhamail: > I would remove these comments

Done ccc94e1

configuration/parse.go
bhamail: > I would remove these comments, too

Done ccc94e1

customerrors/error.go
bhamail: Good point. Changed to "".  ccc94e1

main_test.go
bhamail: > I am always SUPER wary of testing private methods. Do you think we really need to do this, versus an Integration test via dogfooding, etc... that tests more "public" expected behavior?

I felt naked trying to change all the `os.Exit` calls with neither public IT's, nor unit tests of the code being changed. I therefore added some very specific unit tests to enable the refactoring at hand. I agree that IT's at a higher level would be nice to have, but I also believe this is a "chicken and the egg" problem - can't write the high level IT's with all those pesky `os.Exit` calls. Can't do true refactoring (where "refactoring" means making changes that can be proven not to break existing behavior) without unit tests or IT's.

So, I'd be happy to see unit tests be replaced by higher level IT's, which should hopefully be easier to do once the `os.Exit` beast is slain.

main.go
```diff
@@ -181,18 +225,24 @@ func processIQConfig(config configuration.IqConfiguration) {
 	if config.Application == "" {
 		LogLady.Info("No application specified, printing usage and exiting clean")
```
bhamail, line 226: Should a missing "application" arg really exist "clean"? Feels like maybe it should be an error code return value? (Which I added below, but I just noticed the log comment above).

main_test.go
DarthHater: I'll see if I can noodle out a proper IT framework. I agree it would be super useful.

main.go
DarthHater: I think you are right, feel free to go with what makes sense!

bhamail: Done f369b29

audit/csvformatter.go
DarthHater: Maybe an anon function in the function like:

```
	writeAndCheckError = func(w *csv.Writer, value []string) {
		if err = f.write(w, value); err != nil {
			return nil, err
		}
	}
```

Haven't tested it myself but that MIGHT work. Just not sure if it will return in the function itself or return to the block where you call it.

bhamail:
no objections, and some implicit approval, so merging this puppy in a bit.

bhamail:
CAPS LOCK FRIDAY!

PR #138, opened 2020-04-18
Some additional SBOM types to support upstream projects

DarthHater:
For some additional projects (Ahab, Cheque, Hashbrowns), I figured we could extend Nancy to have the core cyclonedx functionality necessary to support additional use cases for generating different types of SBOMs

* Adds a SBOM processor for a slice of `types.Sha1SBOM` (supports Hashbrowns)
* Adds a SBOM processor for a slice of `packageurl.PackageURL` (Ahab, Cheque, and likely Nancy)
* Adds tests for verifying these
* Adds `group` to SBOM output as that was missing previously to packageurl and purl processors

cc @bhamail / @DarthHater

### DIFF
--- a/cyclonedx/cyclonedx.go
+++ b/cyclonedx/cyclonedx.go
@@ -37,11 +37,58 @@ func ProcessPurlsIntoSBOM(results []types.Coordinate) string {
 	return processPurlsIntoSBOMSchema1_1(results)
 }
 
+// SBOMFromPackageURLs will take a slice of packageurl.PackageURL and convert them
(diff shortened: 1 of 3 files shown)


fitzoh:
>https://github.com/sonatype-nexus-community/go-sona-types

nice

DarthHater:
@zendern @fitzoh I am not against that at all. We've got enough tools consuming these things where that would make sense! I'll merge this for now but we can make that a thing for `1.0.0`, rip out the common stuff.

PR #141, opened 2020-05-01
Use v for Version if it's IQ, none if it's OSS Index

DarthHater:
Basically, OSS Index and IQ keep track of versions very differently. OSS Index has versions without the `v` before it, and IQ has it with `v` before it. This makes hybrid approaches slightly interesting.

* Use the method to get purls that OSS Index uses, before we call IQ
* In the CycloneDX generator, if we notice a `v` is missing before a version, add it!
* Updates test for new behavior

cc @bhamail / @DarthHater

### DIFF
--- a/cyclonedx/cyclonedx.go
+++ b/cyclonedx/cyclonedx.go
@@ -19,6 +19,8 @@ package cyclonedx
 
 import (
 	"encoding/xml"
+	"fmt"
+	"strings"
 
 	"github.com/sonatype-nexus-community/nancy/customerrors"
 
@@ -96,6 +98,11 @@ func processPurlsIntoSBOMSchema1_1(results []types.Coordinate) string {
 		if err != nil {
 			_ = customerrors.NewErrorExitPrintHelp(err, "Error parsing purl from given coordinate")
 			return ""
+		}
+
+		// IQ requires a v before versions, so add one if it doesn't exist
+		if !strings.HasPrefix(purl.Version, "v") {
+			purl.Version = fmt.Sprintf("v%s", purl.Version)
 		}
 
 		component := types.Component{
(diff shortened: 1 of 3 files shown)


cyclonedx/cyclonedx.go
```diff
@@ -98,6 +100,11 @@ func processPurlsIntoSBOMSchema1_1(results []types.Coordinate) string {
 			return ""
 		}
 
+		// IQ requires a v before versions, so add one if it doesn't exist
+		if !strings.HasPrefix(purl.Version, "v") {
```
bhamail, line 104: I was scratching my head a while until I realized the cyclonedx code is ONLY called for IQ. Not sure I see a good way to make that clearer, via packaging or some other way - thought about if `cyclonedx*.go` should move under the `iq` package, but not sure....more asking the question.

main.go
```diff
@@ -293,7 +293,7 @@ func doStdInAndParseForIQ(config configuration.IqConfiguration) (err error) {
 		"projectList": mod.ProjectList,
 	}).Debug("Obtained project list")
 
-	var purls = mod.ExtractPurlsFromManifestForIQ()
+	var purls = mod.ExtractPurlsFromManifest()
```
bhamail, line 296: Similar thought about separating concerns of IQ vs. OSSSI processing out the `main.go` code. If only that guy working on the Cobra command stuff would hurry up. ;)

cyclonedx/cyclonedx.go
DarthHater: It likely makes sense to just create a new project like @fitzoh or @zendern suggested: `go-sona-types`, and have sort of "common" functionality in that one.

PR #146, opened 2020-05-23
main: remove figure

greut:
If it's fun in the Spring world, not much in the cli one. _my 2 cents_

Capture d’écran du 2020-05-23 11-22-25

* remove the figure

cc @bhamail / @DarthHater


DarthHater:
I think there isn't really a need for this, you can keep it from showing up with `--quiet`. The reason for it being there is for some minimal branding related to the tool. Can you expand maybe a bit more on why you want to get rid of it?

bhamail:
FWIW, the new default mode will be "quiet", so I think we can close this if you've no objections.

PR #148, opened 2020-05-24
indenting package string cleanly independent of package index

MichelKazi:
* Imported tabwriter to align package string independent of package index

**BEFORE**
image

**AFTER**
image

cc @bhamail / @DarthHater

### DIFF
--- a/audit/auditlogtextformatter.go
+++ b/audit/auditlogtextformatter.go
@@ -22,6 +22,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/jedib0t/go-pretty/v6/text"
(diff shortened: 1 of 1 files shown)


audit/auditlogtextformatter.go
```diff
@@ -22,6 +22,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"text/tabwriter"
```
DarthHater, line 25: Nice! Love seeing you use the standard lib!

audit/auditlogtextformatter.go
```diff
@@ -47,9 +48,13 @@ type AuditLogTextFormatter struct {
 }
 
 func logPackage(sb *strings.Builder, noColor bool, idx int, packageCount int, coordinate types.Coordinate) {
+	w := tabwriter.NewWriter(sb, 9, 3, 0, '\t', 0)
```
DarthHater, line 51: You are using this `tabwriter` in three different places, and setting it up identically. I would probably just do the setup in `Format` and then pass a pointer to it to the methods that need it (sorta like `sb *strings.Builder` above). All of those functions are private so you can change the contract as much as you want!

MichelKazi: Oh sick! I was worried about whether I'd break something by doing that

DarthHater: Naw, we (rather @zendern ) did a great job of making sure the right stuff was private so we could do whatever we wanted.

PR #149, opened 2020-05-27
Start using go-sona-types

DarthHater:
This is a PR that is mostly for @bhamail as the continued quest to Nancy `1.0.0` marches on, this PR starts using the `go-sona-types` project, and rips out the following packages:

- `cyclonedx`
- `iq`
- `ossindex`
- `useragent`

The intent here is to lighten up `nancy` quite a bit, and make it easier for other projects to consume these useful functions, without dragging in EVERY bit of `nancy`'s dependencies, which will grow significantly when using `cobra`. 

* Switches `main.go` and other files to use `go-sona-types`
* Deletes old packages

So far I've tested this against OSS Index and IQ and it's working swimmingly, I fully expect some more work but the initial results are awesome.

cc @bhamail / @DarthHater


DarthHater:
Closing in favor of #158

PR #150, opened 2020-06-10
GO LIST JSON? WHY NOT!?

### LINKED ISSUE #142: Handle replace (old version to other version)

In some upstream testing of `hashbrowns`, I noticed that in your go.mod when you:

```
replace github.com/gorilla/websocket => github.com/gorilla/websocket v1.4.2
```

go list -m all now shows:

```
github.com/gorilla/websocket v1.4.0 => github.com/gorilla/websocket v1.4.2
```

I have severe doubts that `nancy` handles this properly as of this moment, so you'd report on the old version, rather than the replaced one

Basically just keeping people up to date on what they are actually using!

Not entirely sure. There is a hacky way to do this, and then probably a less hacky way. One idea is to look at the structs, etc... that `go mod` itself uses, and see if we can leverage those for parsing the output similar to what we do with `dep`.

* **Anything else?**

Fun?!

cc @bhamail / @DarthHater

DarthHater:
This PR handles:

`go list -m all` and `go list -json -m all` commands piped in to Nancy

I went with this approach as the Go Mod list command stuff is all internal. I've copied the struct for `Module` however, and trimmed it down for now to the fields we likely will ever need.

The approach I took was to:

- Read the stdIn into a `[]byte`, as I don't anticipate it being TOO big (I tried at first to do this by buffering it, but the code was getting extremely complicated for something that should never be like, massive in memory footprint)
- Attempt to read from the `[]byte` as json. If it errors, break the loop, if it doesn't, return everything, GREAT!
- If it exits the loop due to an error (not EOF), I attempt to read it as if it was just a regular list like what we had before
- The function that reads the parsed JSON can handle mods that were replaced now, giving us the accurate version being used

This new method `GoParseAgnostic` takes an `io.Reader` as well, so that we can test it a bit easier.

* Switches `main.go` to use `GoListAgnostic`
* Implements `func GoListAgnostic(stdIn io.Reader) (deps types.ProjectList, err error)`
* Implements a private function to parse a `GoListModule` struct, which has a signature of `func modToProjectList(mod types.GoListModule) (dep types.Projects, err error)`

* Fixes #142 

cc @bhamail / @DarthHater


parse/parse.go
```diff
@@ -39,6 +43,68 @@ func GoList(stdIn *bufio.Scanner) (deps types.ProjectList, err error) {
...
+func modToProjectList(mod types.GoListModule) (dep types.Projects, err error) {
+	if mod.Replace != nil {
+		if mod.Replace.Version == "" {
+			err = &NoVersionError{err: fmt.Errorf("No version found for mod")}
+			return
+		}
+		dep.Name = mod.Replace.Path
+		dep.Version = mod.Replace.Version
+		return
+	}
```
zendern, line 98: Since #142 called this out as an issue i think we should add a test around it to make sure it is working as expected.

DarthHater:
@zendern , README updated!

parse/parse.go
DarthHater: I have one! https://github.com/sonatype-nexus-community/nancy/pull/150/files#diff-7b16ed199b45ba365ab40918a65bbb2bR64-R78

zendern: My bad i should have been more clear :) that just handles `replace` for `go list` and not `go list -json`. We should probably have a replace test for the json path too. 

```
{
        "Path": "github.com/gorilla/websocket",
        "Version": "v1.4.0",
        "Replace": {
                "Path": "github.com/gorilla/websocket",
                "Version": "v1.4.2",
                "Time": "2020-03-19T17:50:51Z",
                "GoMod": "/Users/ME/go/pkg/mod/cache/download/github.com/gorilla/websocket/@v/v1.4.2.mod"
        },
        "Update": {
                "Path": "github.com/gorilla/websocket",
                "Version": "v1.4.2",
                "Time": "2020-03-19T17:50:51Z"
        },
        "Indirect": true,
        "GoMod": "/Users/ME/go/pkg/mod/cache/download/github.com/gorilla/websocket/@v/v1.4.2.mod"
}
```

One of those testdata files but with that and maybe another in it.

DarthHater: In: 64079e0

PR #151, opened 2020-06-10
Add log command

protoworlock69:
Add log command that will output the Nancy Log to the console

* added command and implementation in main.go
cc @bhamail / @DarthHater

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -29,6 +30,7 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	"github.com/sonatype-nexus-community/nancy/logger"
 	. "github.com/sonatype-nexus-community/nancy/logger"
 	"github.com/sonatype-nexus-community/nancy/types"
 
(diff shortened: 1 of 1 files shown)


DarthHater:
This is great, Dahlia!

We need to fix the logger a bit before we can move this forward. Unfortunately the way I wrote the logger makes it wipe the file a bit TOO quickly (when nancy starts up), I can work on that with you and @TheDoroTheExploro pretty soon!

main.go
```diff
@@ -67,6 +71,12 @@ func main() {
 	}
 }
 
+func printLog() {
+	file, _ := os.Open(logger.GetLogFileLocation())
```
DarthHater, line 75: We should check the error here, so an improvement to make would be:

```
file, err := os.Open(logger.GetLogFileLocation())
if err != nil {
     fmt.Println(err)
}
```

main.go
```diff
@@ -67,6 +71,12 @@ func main() {
...
+	b, _ := ioutil.ReadAll(file)
```
DarthHater, line 76: We should check the error here, so an improvement to make would be:

```
b, err := ioutil.ReadAll(file)
if err != nil {
     fmt.Println(err)
}
```

main.go
zendern: I concur on both of these probably wouldn't hurt to also have some sort of user friendly message along with the raw error that happened. 

`The log you were looking for wasn't on the pile with the others ¯\_(ツ)_/¯` :)

PR #155, opened 2020-06-25
Goodbye go.sum old friend

### LINKED ISSUE #69: Remove go.sum option
Only support the new authoritative way in go to get a list of dependencies. `go list -m all` should be the way, the path, the truth. So allowing for using `go.sum` should no longer be needed. 

To not support 2 ways to deal with go mod. `go list -m all` is the authoritative way to do the thing so lets do that.

Remove some code :) 

* **Anything else?**
When we do this its probably a major bump of nancy as we are removing functionality 

cc @bhamail / @DarthHater

zendern:
You were fine to start but newer better things came along. We will never forget your
contributions to the cause.

* Removes anything related to go.sum

(If there are changes to user behavior in general, please make sure to
update the docs, as well)
^^^ I think i got these and replaced them with there `go list -json -m all` equivalent. 

* Fixes #69 

### Should not be merged yet since should be a 1.0 thing. 
Also maybe we cut a branch for that and start stashing these PRs in it?? Maybe even start to cut Release Candidates for 1.0??

cc @bhamail / @DarthHater


bhamail:
I'm pretty sure this work has been merged into #161. Holler if not.

bhamail:
fixed by #161

PR #156, opened 2020-06-27
Update docker usage

kishaningithub:
Add --rm so that the container gets destroyed after nancy completes its execution


DarthHater:
Sounds simple enough! Was having the old container around causing issues for you @kishaningithub ?

@zendern @MichelKazi @bhamail if any of you want to take a look, this is really simple and I think can get merged!

Issue #160, opened 2020-08-05
Un-publify packages not designed for public consumption.


bhamail:
Make some packages "internal" before 1.0.0.

e.g. `logger`, `audit`, `configuration`, `packages`

maybe leave `parse` as public

cc @bhamail / @DarthHater

bhamail:
Fixed by #163

PR #161, opened 2020-08-05
Prepare 1.0.0 - all the stuff we want to fix before the official initial release

bhamail:
attempting to include most of the breaking changes we know about. 

include @zendern changes to remove go.sum.

Holler with stuff we missed. 

e.g. 
#160 - Un-publify packages not designed for public consumption

cc @bhamail / @DarthHater


README.md
```diff
@@ -169,8 +169,8 @@ This can be handy for testing your account out, or if you want to override your
 
 You can run `nancy` in a quiet manner, only getting back a list of vulnerable components by running:
 
-* `./nancy -quiet /path/to/your/Gopkg.lock `
-* `./nancy -quiet /path/to/your/go.sum `
+* `./nancy -quiet /path/to/your/Gopkg.lock`
```
DarthHater, line 172: Pretty sure it's `-p` now for path yeah?

README.md
```diff
@@ -183,14 +183,14 @@ We support exclusion of vulnerability either by CVE-ID (ex: `CVE-2018-20303`) or
 
 ##### Via CLI flag
 * `./nancy -exclude-vulnerability CVE-789,bcb0c38d-0d35-44ee-b7a7-8f77183d1ae2 /path/to/your/Gopkg.lock`
-* `./nancy -exclude-vulnerability CVE-789,bcb0c38d-0d35-44ee-b7a7-8f77183d1ae2 /path/to/your/go.sum`
+* `go list -json -m all | ./nancy -exclude-vulnerability CVE-789,bcb0c38d-0d35-44ee-b7a7-8f77183d1ae2`
```
DarthHater, line 186: Likely `--exclude-vulnerability` I think

README.md
```diff
@@ -183,14 +183,14 @@ We support exclusion of vulnerability either by CVE-ID (ex: `CVE-2018-20303`) or
...
 
 ##### Via file
 By default if a file named `.nancy-ignore` exists in the same directory that nancy is run it will use it, will no other options need to be passed.
 
 If you would like to define the path to the file you can use the following
 * `./nancy -exclude-vulnerability-file=/path/to/your/exclude-file /path/to/your/Gopkg.lock`
-* `./nancy -exclude-vulnerability-file=/path/to/your/exclude-file /path/to/your/go.sum`  
+* `go list -json -m all | ./nancy -exclude-vulnerability-file=/path/to/your/exclude-file`
```
DarthHater, line 193: Likely `--exclude-vulnerabiliity-file` now

audit/csvformatter_test.go
```diff
@@ -40,8 +40,7 @@ func TestCsvOutputWhenQuiet(t *testing.T) {
 	}
 	entry := Entry{Data: data}
 
-	quiet := true
-	formatter := CsvFormatter{Quiet: &quiet}
+	formatter := CsvFormatter{Quiet: true}
```
DarthHater, line 43: Might be an unintended consequence, not sure if we want to pass quiet or true all the time (worth a look)

cmd/iq.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if fileExists(cfgFileToCheck) {
+		// 'merge' IQ config here, since we also need OSSI config, and load order is not guaranteed
+		if err := viper.MergeInConfig(); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func auditWithIQServer(purls []string, applicationID string) error {
```
DarthHater, line 204: Unrelated to this (just making an observation), for `Options` on the IQServer `go-sona-types` struct, I created an `Application` string, which I had sort of intended to allow us to set that in the `Options` rather than pass it all over the place. HOWEVER, I never used it in `AuditPackages` etc..., so you know, wishful thinking. I can likely get back to `go-sona-types` and implement it so that it's used.

cmd/iq_test.go
```diff
@@ -0,0 +1,273 @@
...
+	defer func() {
+		resetOSSIConfigFile()
+	}()
+
+	setupIQConfigFile(t, tempDir)
+	defer func() {
+		resetIQConfigFile()
+	}()
+
+	credentials := fmt.Sprintf("%s: %s\n", configuration.YamlKeyIQUsername, "iqUsernameValue") +
```
DarthHater, line 76: You COULD do:

```
credentials := fmt.Sprintf("%s: %s\n%s: %s\n%s: %s\n", 
configuration.YamlKeyIQUsername, 
"iqUsernameValue", 
configuration.YamlKeyIQToken, 
"iqTokenValue", 
configuration.YamlKeyIQServer, "iqServerValue",
)
```

I THINK

cmd/iq_test.go
```diff
@@ -0,0 +1,273 @@
...
+	setupTestOSSIConfigFileValues(t, tempDir)
+	defer func() {
+		resetOSSIConfigFile()
+	}()
+
+	setupIQConfigFile(t, tempDir)
+	defer func() {
+		resetIQConfigFile()
+	}()
+	credentials := fmt.Sprintf("%s: %s\n", configuration.YamlKeyIQUsername, "iqUsernameValue") +
```
DarthHater, line 110: You COULD do:

```
credentials := fmt.Sprintf("%s: %s\n%s: %s\n%s: %s\n", 
configuration.YamlKeyIQUsername, 
"iqUsernameValue", 
configuration.YamlKeyIQToken, 
"iqTokenValue", 
configuration.YamlKeyIQServer, "iqServerValue",
)
```
I THINK

README.md
bhamail: > Pretty sure it's `-p` now for path yeah?

Good catch! I also swapped things to document default==quiet, and new --loud flag.  cfdd7de

README.md
bhamail: > Likely `--exclude-vulnerability` I think

Yes! Also fixed some other inconsistencies in: 25c11c1

README.md
bhamail: Thanks! 25c11c1

audit/csvformatter_test.go
bhamail: > Might be an unintended consequence, not sure if we want to pass quiet or true all the time (worth a look)

The `quiet` default is provided by the Cobra command. We could add a new test of the `loud` mode if desired.
Also, the old `&quiet` reference went away went away.

Sorry if I've totally missed the point.

cmd/iq.go
bhamail: > Unrelated to this (just making an observation), for `Options` on the IQServer `go-sona-types` struct, I created an `Application` string, which I had sort of intended to allow us to set that in the `Options` rather than pass it all over the place. HOWEVER, I never used it in `AuditPackages` etc..., so you know, wishful thinking. I can likely get back to `go-sona-types` and implement it so that it's used.

👍 Might be "Issue worthy" to help remember todo it.

cmd/iq_test.go
bhamail: Yes! Thank you again: 5a4c14e

cmd/iq_test.go
bhamail: Yes! Thank you!

bhamail:
After looking at how @zendern used the `ahab chase` pattern in ahab, I'm now wondering if nancy should use a `nancy slueth` pattern? I would move the ossi logic into a new Cobra sub-command (`slueth`). This would leave the `root` command very simple (as it is in ahab). Or is that too punny, and would `nancy ossi` be clearer?

Or is this change just too much for this already huuuge PR?

(I think @DarthHater suggested this sort of change before, but I'm only now grokking the idea)

PR #165, opened 2020-08-13
separate `sleuth` command for 'oss index'

bhamail:
This PR creates a new sub command named `sleuth` for running nancy against OSS Index.

There is more work to do regarding updating cmd examples and README.md, but I wanted to put this out there to get initial reactions.

cc @bhamail / @DarthHater

### DIFF
--- a/cmd/iq_test.go
+++ b/cmd/iq_test.go
@@ -53,7 +53,7 @@ func TestIqCommandPathInvalidName(t *testing.T) {
 	defer func() {
 		configOssi = origConfig
 	}()
-	// not sure why calling executeCommand() fails as part of test suite, but is fine individually.
+	// TODO debug side effects. calling executeCommand() fails as part of test suite, but is fine when run individually.
 	//_, err := executeCommand(rootCmd, iqCmd.Use, "--path", "invalidPath", "-a", "appId")
 	configOssi = types.Configuration{Path: "invalidPath"}
 	err := doIQ(iqCmd, []string{})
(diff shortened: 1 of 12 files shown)


DarthHater:
I am pro this, overall. It certainly makes the code a lot cleaner and easier to follow (I think), so feel free to keep moving on it.

cmd/iq_test.go
```diff
@@ -35,18 +36,46 @@ import (
...
 	assert.Nil(t, err)
 }
 
+func TestIqCommandPathInvalidName(t *testing.T) {
+	origConfig := configOssi
+	defer func() {
+		configOssi = origConfig
+	}()
+	// not sure why calling executeCommand() fails as part of test suite, but is fine individually.
+	//_, err := executeCommand(rootCmd, iqCmd.Use, "--path", "invalidPath", "-a", "appId")
```
zendern, line 57: So these are a fix it later thing??

cmd/iq_test.go
```diff
@@ -35,18 +36,46 @@ import (
...
+	assert.Contains(t, err.Error(), fmt.Sprintf("invalid path value. must point to '%s' file. path: ", GopkgLockFilename))
+}
+
+func TestIqCommandPathInvalidFile(t *testing.T) {
+	origConfig := configOssi
+	defer func() {
+		configOssi = origConfig
+	}()
+	// not sure why calling executeCommand() fails as part of test suite, but is fine individually.
+	//_, err := executeCommand(rootCmd, iqCmd.Use, "--path", GopkgLockFilename, "-a", "appId")
```
zendern, line 71: So these are a fix it later thing??

cmd/iq_test.go
bhamail: > So these are a fix it later thing??

Yes, or at least an "understand it better later" thing. Added TODO comments. cbcbbe3 

I'm pretty sure this is an example of the dangers of using shared state - meaning the `configOssi` is essentially global, and testing gets wonky quickly because of the side effects of manipulating `configOssi`. Combine that with some of the essentially global state of Cobra/Viper initialization, and complexity abounds.

I would prefer these tests to work using `executeCommand()` because that would verify more of the Cobra stuff, but I spent enough time trying to figure out why the `executeCommand()` flavor works when run alone, but causes failures when run in a test suite with other tests.

cmd/iq_test.go
bhamail: > So these are a fix it later thing??

Yes, or at least an "understand it better later" thing. Added TODO comments. cbcbbe3

bhamail:
If not objections, I'll merge this puppy, now that the build is passing.

PR #175, opened 2020-08-25
Use goreleaser to build native linux installers

bhamail:
Because goreleaser rocks! These config changes will build native linux installers: .deb, .rpm, and .apk

Probably need more work to deal with signing, etc, but this is a first step.

cc @bhamail / @DarthHater

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -110,3 +110,13 @@ dockers:
     - "sonatypecommunity/nancy:{{ .Tag }}-alpine"
     - "sonatypecommunity/nancy:v{{ .Major }}-alpine"
     - "sonatypecommunity/nancy:v{{ .Major }}.{{ .Minor }}-alpine"
+
+nfpms:
+  -
+    vendor: sonatype-nexus-community
+    homepage: https://github.com/sonatype-nexus-community/nancy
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -110,3 +110,13 @@ dockers:
...
     - "sonatypecommunity/nancy:v{{ .Major }}-alpine"
     - "sonatypecommunity/nancy:v{{ .Major }}.{{ .Minor }}-alpine"
+
+nfpms:
+  -
+    vendor: sonatype-nexus-community
+    homepage: https://github.com/sonatype-nexus-community/nancy
+    description: "A tool to check for vulnerabilities in your Golang dependencies, powered by Sonatype OSS Index"
+    formats:
+      - apk
```
DarthHater, line 120: So this will just build them and attach them to the release, yeah? Do we have a plan for distributing them further? Just curious!

zendern: I think we should that would be neat and same goes for ahab :)

zendern: I just opened #177 to address that.

zendern: and this one for ahab :) https://github.com/sonatype-nexus-community/ahab/issues/38

PR #185, opened 2020-09-11
Update license headers

tneer:
Update the license header on a few files that weren't passing the internal build license header check.

* Add/tweak license headers

cc @bhamail / @DarthHater


bhamail (approved):
Thanks @tneer! BTW, I've added an issue to help track adding a license header check: #186

Issue #189, opened 2020-09-16
Incorrect reporting for vulnerability for golang.org/x/net for latest updates


harshavardhana:
Thanks for creating an issue! Please fill out this form so we can be
sure to have all the information we need, and to minimize back and forth.

* **What are you trying to do?**
Trying to fix a vulnerability report

```
[1/1]	pkg:golang/golang.org/x/net@0.0.0-20200904194848-62affa334b73
6 known vulnerabilities affecting installed version 
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ [CVE-2018-17075]  Improper Input Validation                                                                                                            ┃
┣━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Description        ┃ The html package (aka x/net/html) before 2018-07-13 in Go mishandles "in                                                          ┃
┃                    ┃ frameset" insertion mode, leading to a "panic: runtime error" for                                                                 ┃
┃                    ┃ html.Parse of <template><object>, <template><applet>, or                                                                          ┃
┃                    ┃ <template><marquee>. This is related to HTMLTreeBuilder.cpp in WebKit.                                                            ┃
┣━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ OSS Index ID       ┃ ae060e1b-dedb-41ac-8e61-235d179157b8                                                                                              ┃
[...]

PR #190, opened 2020-09-22
handle IQ policy Warning.

bhamail:
Print a message with a hint about policy Warnings from IQ.

Discovered while setting up IQ dog-fooding

cc @bhamail / @DarthHater

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,8 +43,14 @@ test: build
 	$(GOTEST) -v ./... 2>&1
 
 integration-test: build
+	# temporary workaround, remove next line when x/net false positive is fixed
+	echo 'CVE-2018-17142\nCVE-2018-17846\nCVE-2018-17143\nCVE-2018-17847\nCVE-2018-17848' > packages/testdata/.nancy-ignore
 	cd packages/testdata && GOPATH=. ../../$(BINARY_NAME) sleuth -p Gopkg.lock && cd -
+	# temporary workaround, remove next line when x/net false positive is fixed
+	mv packages/testdata/.nancy-ignore .
 	go list -json -m all | ./$(BINARY_NAME) sleuth
 	go list -m all | ./$(BINARY_NAME) sleuth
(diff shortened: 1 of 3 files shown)


internal/cmd/iq.go
```diff
@@ -249,8 +249,13 @@ func auditWithIQServer(purls []string) error {
 
 	if res.PolicyAction != "Failure" {
 		logLady.WithField("res", res).Debug("Successful in communicating with IQ Server")
-		fmt.Println("Wonderbar! No policy violations reported for this audit!")
-		fmt.Println("Report URL: ", res.ReportHTMLURL)
+		if res.PolicyAction == "Warning" {
```
DarthHater, line 252: Personally, I'd switch to a `switch` on this, over an `if else`, I generally do that any time I'm over two values.

bhamail: Thanks for the reminder! 8edd102

Makefile
```diff
@@ -43,8 +43,14 @@ test: build
 	$(GOTEST) -v ./... 2>&1
 
 integration-test: build
+	# temporary workaround, remove next line when x/net false positive is fixed
+	echo 'CVE-2018-17142\nCVE-2018-17846\nCVE-2018-17143\nCVE-2018-17847\nCVE-2018-17848' > packages/testdata/.nancy-ignore
```
DarthHater, line 47: I would add a `CVN-111 until=2021-01-01` type thing to this, so they are only ever ignored for like a week or two. We really don't want to have these ignored forever if they become applicable.

DarthHater (changes requested):
Likely want to only ignore the CVEs until a specific date, asked for a change related to that

Makefile
bhamail: Yes! Done: e313db8  Also added comment to Issue #189

DarthHater (approved):

PR #85, opened 2020-03-01
gorelease: build image suitable for CI

greut:
Create a `nancy:ci` Docker image that is suitable for Continuous integration such as Gitlab CI (where a sh is required)

This pull request makes the following changes:
* remove the `v` duplicate in the archive name in `--snapshot` mode.
* creates a second image for ci usage `nancy:ci`
* stop running the update/upgrade inside docker but pull the latest image

cc @bhamail / @DarthHater

PR #195, opened 2020-10-07
Docker tests ftw

### LINKED ISSUE #119: Integration/Dogfood test for our docker image
We have really good dogfood tests for nancy on itself for the binary build but we do not have that in place for the Docker image. 

Doing the same dogfood test for the docker image to make sure we dont break it

Probably need to build the Dockerfile at build time (even though it will happen again as part of goreleaser) and then run it and make sure it does what we expect. 

* **Anything else?**

cc @bhamail / @DarthHater

zendern:
* yyyuummmm dogfood on a boat
* little bit of cleanup as well

* Fixes #119 

cc @bhamail / @DarthHater


.circleci/config.yml
```diff
@@ -41,6 +41,8 @@ jobs:
       - run: make ci-lint
       - run: make test | go-junit-report > $TEST_RESULTS/gotest/report.xml
       - run: make integration-test
+      - setup_remote_docker
```
DarthHater, line 44: Generally you probably want to do this as the first step in the build? I'd move it higher, as a result. Mostly because it's environment config?

Makefile
```diff
@@ -43,14 +43,38 @@ test: build
...
+	mkdir -p dist
+	docker build . -f Dockerfile.alpine -t sonatypecommunity/nancy:alpine-integration-test
+	# create file, volume mount to simulate, ci run of the container and things just happening inside the container instead of passing output to the container directly
+	go list -json -m all > dist/deps.out
+	echo "cat /tmp/dist/deps.out | nancy sleuth" > dist/ci.sh
+	chmod +x dist/ci.sh
+	# run the container....using cat with no params keeps it running
+	docker run --name alpine-integration-test -td sonatypecommunity/nancy:alpine-integration-test cat
+	# copy the code as if it was actually in the "ci" container.. doing this cause circleci cant actually mount volumes
+	docker cp . alpine-integration-test:/tmp
```
DarthHater, line 72: A question here, is why not in `Dockerfile.alpine` do a `COPY dist dist` or something akin, and control it via a build arg? Was it just not wanting to modify the Dockerfile itself?

DarthHater: I'm not suggesting that approach, just you know, unpacking this approach.

Makefile
```diff
@@ -43,14 +43,38 @@ test: build
...
+
+docker-alpine-integration-test: build-linux
+	mkdir -p dist
+	docker build . -f Dockerfile.alpine -t sonatypecommunity/nancy:alpine-integration-test
+	# create file, volume mount to simulate, ci run of the container and things just happening inside the container instead of passing output to the container directly
+	go list -json -m all > dist/deps.out
+	echo "cat /tmp/dist/deps.out | nancy sleuth" > dist/ci.sh
+	chmod +x dist/ci.sh
+	# run the container....using cat with no params keeps it running
+	docker run --name alpine-integration-test -td sonatypecommunity/nancy:alpine-integration-test cat
```
DarthHater, line 70: Why does it need to be running? I just noticed in the `Dockerfile.alpine` that we are running `nancy` and not `nancy sleuth`, by the way.

I just am curious, shouldn't the `docker exec` be sufficient?

.circleci/config.yml
zendern: Done in 4a9c74d

Makefile
zendern: So yah didnt want to make the docker file bigger and package the dist directory in it. This is just an attempt to simulate as if you did the following. 

```
build:
    docker:
      - image: sonatypecommunity/nancy:alpine
    working_directory: /go/src/github.com/sonatype-nexus-community/nancy
    steps:
       - nancy (sleuth|iq) ...
```

Makefile
zendern: So it has to be running to be able to copy files to it from what i can tell from here. 

https://docs.docker.com/engine/reference/commandline/cp/

Since the container has no ENTRYPOINT by design (see here https://github.com/sonatype-nexus-community/nancy/pull/85) it will just automatically stop if you just run it.

Honestly we might not even need `COMMAND` defined at all the way the container is used. In the other docker container that allows you to not "install" nancy we dont want sleuth there b/c we want to give you the ability to do `sleuth` or `iq`. 

Example :
`go list -json -m all | docker run --rm -i sonatypecommunity/nancy:latest sleuth`

bhamail:
This is waaay cool! I was seeing some errors when calling `docker build` on macos.

Found this: https://discuss.circleci.com/t/circleci-build-local-machine-fails-when-using-setup-remote-docker/15487
and: https://stackoverflow.com/questions/45796661/docker-permission-denied-while-trying-to-connect-to-docker-daemon-with-local-ci
I'm working on a diff that fixes things 'nuff to allow my local CI build to succeed - basically by running 'sudo docker' when local.

zendern:
> This is waaay cool! I was seeing some errors when calling `docker build` on macos.
> 
> Found this: https://discuss.circleci.com/t/circleci-build-local-machine-fails-when-using-setup-remote-docker/15487
> and: https://stackoverflow.com/questions/45796661/docker-permission-denied-while-trying-to-connect-to-docker-daemon-with-local-ci
> I'm working on a diff that fixes things 'nuff to allow my local CI build to succeed - basically by running 'sudo docker' when local.
> 
> For reference, here's the command I'm using to run a local CI build:
> 
> ```
> circleci config process .circleci/config.yml > .circleci/local-config.yml && circleci local execute -c .circleci/local-config.yml --job 'build'
> ```

Alright finally got around to looking at this. But take a peek at this commit and try to run it again for me please. 2e4c7a0 Details of what is going on are in that commit.

bhamail:
@zendern Oh, snaaaap! Nicely done. Works like a champ! Merging it right away!!!! Thanks fo much!

bhamail (approved):
Very nice!

PR #198, opened 2020-10-08
Cleanup Jenkinsfile to use Makefile

### LINKED ISSUE #120: JenkinsFile using Makefile
Cleanup a little duplication

Whats in the .circleci/config.yml and JenkinsFile are pretty similar. And some of it already is in the Makefile. 

So to keep everything more in sync switch over parts that are possible and maybe pull more into the Makefile. 

* **Anything else?**

cc @bhamail / @DarthHater

zendern:
* Fixes #120

Just simply making things consistent where we can. CircleCi uses Makefile and we use it locally. So we should use it in the Jenkinsfile as well. 

cc @bhamail / @DarthHater

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,11 +23,10 @@ dockerizedBuildPipeline(
   },
   buildAndTest: {
     sh '''
-    go mod download
-    go mod tidy
     go get -u github.com/jstemmer/go-junit-report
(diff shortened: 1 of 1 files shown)


bhamail (approved):
+1, will fire off Jenkins build after merging.

PR #208, opened 2021-01-08
Update check

### LINKED ISSUE #140: Provide option for nancy to auto-update
Would be neat to have nancy (optionally) automatically update itself.

Could take approach similar to that used by CircleCi client. Libs to consider:

https://github.com/rhysd/go-github-selfupdate

https://godoc.org/github.com/tj/go-update

Should use GitHub releases. 

Be aware of handling "brew" installed flavor - I think the CircleCI client handles this.

cc @bhamail / @DarthHater

bhamail:
Automatic check for new releases. Prompt user when new version is available.

The automatic check will only occur once every 28 hours.
The timestamp of the lookup is stored in an app specific yml file:
`~/.ossindex/.nancy-config/update_check.yml`

This PR also adds an explicit `update` command that can be used to force a check for a new release. e.g.
`nancy update`

In order for the release lookup logic to be happy, I had to change the goreleaser release filename template a little. (see `.goreleaser.yml`)
The related logic in the selfupdater is: https://github.com/rhysd/go-github-selfupdate/blob/master/selfupdate/detect.go#L95. The suffixes used to find a matching release file didn't like the 'version' on the tail end, so I moved it.
Note: I'm guessing something in [homebrew-nancy-tap](https://github.com/sonatype-nexus-community/homebrew-nancy-tap) will need to change to match the new filename pattern (or maybe it'll "just work")?

Once we have some releases with the new filename pattern, I can work on adding "auto update" stuff for certain cases.

* Fixes #140

cc @bhamail / @DarthHater


bhamail:
I will look into merging the goreleaser template changes first. That would allow me to do more testing.

internal/cmd/update.go
```diff
@@ -0,0 +1,108 @@
...
+
+	/*	updateCmd.AddCommand(&cobra.Command{
+			Use:    "check",
+			Hidden: true,
+			Short:  "Check if there are any updates available",
+			RunE: func(_ *cobra.Command, _ []string) error {
+				return updateCLI("", false)
+			},
+		})
+	*/
```
cah-nathanzender, line 37: So why the added `check` here?? Is this the future?? since there is also an `install` commented out below??

internal/cmd/update.go
```diff
@@ -0,0 +1,108 @@
...
+				opts.cfg.SkipUpdateCheck = true
+			},
+			PreRun: func(cmd *cobra.Command, args []string) {
+				opts.args = args
+			},
+			RunE: func(_ *cobra.Command, _ []string) error {
+				return updateCLI(opts)
+			},
+		})
+	*/
```
zendern, line 52: So what are these?? Future state?? If the plan is to have it be `nancy update check` and `nancy update install` we should probably first pass make it actually be `nancy update check` instead of changing the api again probably very soon.

internal/cmd/check.go
```diff
@@ -0,0 +1,73 @@
...
+	updateCheck := &settings.UpdateCheck{
+		LastUpdateCheck: time.Time{},
+	}
+
+	err := updateCheck.Load()
+	if err != nil {
+		return err
+	}
+
+	if update.ShouldCheckForUpdates(updateCheck) {
```
zendern, line 21: This is probably fine....but thinking of how we do CI.... not much gets cached. So once we make this automatic its going to be checking all the time if the underlying docker CI containers always get blown away. We we still want to do it automatically we probably need to add something to the README about caching the ~/.ossindex dir.

internal/cmd/update.go
```diff
@@ -0,0 +1,108 @@
...
+	//updateCmd.PersistentFlags().BoolVar(&opts.dryRun, "check", false, "Check if there are any updates available without installing")
+
+	return updateCmd
+}
+
+func logAndShowMessage(message string) {
+	logLady.Info(message)
+	fmt.Println(message)
```
zendern, line 60: Do we want to respect the quiet flag here??

zendern: Maybe its fine.... realistically this is for humans only right?? Not really a CI thing and probably doesn't hurt to log some more.

buildversion/version.go
```diff
@@ -22,3 +22,12 @@ var (
 	BuildTime    = ""
 	BuildCommit  = ""
 )
+
+// PackageManager defines the package manager which was used to install the CLI.
+// You can override this value using -X flag to the compiler ldflags. This is
+// overridden when we build for Homebrew.
```
zendern, line 28: So this still needs done so that `homebrew` will be set here??

zendern:
```$ ./nancy update
Checking for updates...
You are running 0.0.0
A new release is available (1.0.6)
You can visit the Github releases page for the CLI to manually download and install:
https://github.com/sonatype-nexus-community/nancy/releases```

As per @bhamail ... this is some sample output

File: internal/cmd/update.go (hunk 2 above)
Comment (bhamail) [reply internal/cmd/update.go:52]: > So what are these?? Future state?? If the plan is to have it be `nancy update check` and `nancy update install` we should probably first pass make it actually be `nancy update check` instead of changing the api again probably very soon.

Yeah, far future state. The really nice [CircleCI CLI code](https://github.com/CircleCI-Public/circleci-cli/blob/master/cmd/update.go#L52) I'm basing this off of has some complex (to my brain anyway) logic that sets some flags (`opts...`) on "hidden" commands to affect if/when an "install" is attempted (`dryRun`). I don't think I want to go down the same path. Maybe I'll change my mind if a light bulb goes on for me, but I'm gonna try and keep it simpler.

I kept these commented chunks around mostly to remember to flesh the "install" stuff.

File: internal/cmd/check.go (hunk 3 above)
Comment (bhamail) [reply internal/cmd/check.go:27]: > This is probably fine....but thinking of how we do CI.... not much gets cached. So once we make this automatic its going to be checking all the time if the underlying docker CI containers always get blown away. We we still want to do it automatically we probably need to add something to the README about caching the ~/.ossindex dir.

Great point! I'll try to add some doco about this. I actually like the idea of it printing messages about an updated Nancy version in CI builds. Slightly increasing the chance a human will notice it in the CI logs. Still worth mentioning though.

File: internal/cmd/update.go (hunk 4 above)
Comment (bhamail) [reply internal/cmd/update.go]: I see your thinking, and am still leaning to leaving the "nag" enabled even in "quiet" mode. If somebody complains, we can tweak things.

File: buildversion/version.go (hunk 5 above)
Comment (bhamail) [reply buildversion/version.go:28]: > So this still needs done so that `homebrew` will be set here??

Err, um, yes? This falls into the bucket of "I still don't know exactly how that will all work". :)

**Comment by bhamail** [comment]
Fixed a boog in the autocheck execution path. Some sample output:
```
$ ./nancy 
Checking for updates...
You are running 0.0.0
A new release is available (1.0.6)
You can visit the Github releases page for the CLI to manually download and install:
https://github.com/sonatype-nexus-community/nancy/releases


Usage:
  nancy [flags]
...
```

internal/cmd/check.go
bhamail: Added some docs here: c62fded

internal/cmd/update.go
bhamail: Deleted these comments in 1882f98

Created https://github.com/sonatype-nexus-community/nancy/issues/209 to track future work.

PR #219, opened 2021-02-11
Add a non-versioned "latest" artifact

### LINKED ISSUE #47: Remove version from github release artifact name
Make it even easier to install nancy and stay up to date.

If i wanted to download nancy as part of a build and not have to build from source it should be easy to do and easy to keep up to date. 

Github has an auto redirect if you use the following url that will take you to the latest release.
`https://github.com/sonatype-nexus-community/nancy/releases/download/latest/nancy-darwin.amd64-v0.0.31`

But it doesn't really do much good for us since we include the version number in the artifact name. 

Proposal is to switch it to this. That way we can suggest that people just curl/wget this url and it will give the the latest and greatest nancy binary available. 
`https://github.com/sonatype-nexus-community/nancy/releases/download/latest/nancy-darwin.amd64`

Remove the version number from the release artifact when releasing via travis

cc @bhamail / @DarthHater

bhamail:
There are a number of use cases enabled by producing a "non-versioned" (aka "latest") binary when releasing. This PR adds such binaries (I think...).

* Fixes #47

cc @bhamail / @DarthHater

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -99,7 +99,7 @@ dockers:
     goarch: amd64
     dockerfile: build/Dockerfile.alpine
     binaries:
-    - nancy
+      - nancy
     build_flag_templates:
     - "--pull"
     - "--label=author='DJ Schleen'"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -83,7 +83,7 @@ dockers:
     goos: linux
     goarch: amd64
     dockerfile: Dockerfile.goreleaser
-    binaries:
+    ids:
       - nancy
     build_flag_templates:
     - "--label=author='DJ Schleen'"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -83,7 +88,7 @@ dockers:
     goos: linux
     goarch: amd64
     dockerfile: Dockerfile.goreleaser
-    binaries:
+    ids:
```
DarthHater, line 91: Did something change in goreleaser?

DarthHater (approved):
Other than a tiny question, I see no issue here. Gonna have to test in PRODUCTION!!!

.goreleaser.yml
bhamail: It was nagging about it:
```
• docker images    
            • DEPRECATED: `docker.binaries` should not be used anymore, check https://goreleaser.com/deprecations#dockerbinaries for more info.
            • DEPRECATED: `docker.binaries` should not be used anymore, check https://goreleaser.com/deprecations#dockerbinaries for more info.
```

PR #220, opened 2021-03-10
panic a little less on replaces

### LINKED ISSUE #207: Panics with index out of range error

```
$ git clone git@github.com:ory/kratos.git
$ cd kratos
$ git checkout -b nancy 9ea96120451cf93be42c87146dcb9db02c67631d
$ go list -m all | nancy sleuth -q
Error: runtime error: index out of range [4] with length 4

[... rest of help message ...]
```

* **Anything else?**

cc @bhamail / @DarthHater

cah-nathan-zender:
☝️ hopefully have this happen less 😄 

------------------

When a replace happened to a package that was internal to the project it
blew up this fixes that and validates that the versions should be pulled
from pre-internal project replace and still scanned.

----------------

As the comment says it'll still pull in the old version pre replace and send it along for scanning. Need to determine if that is the right thing to do or not. 

Example 
`github.com/ory/kratos-client-go v0.5.4-alpha.1 => ./internal/httpclient`

Will end up as 
`github.com/ory/kratos-client-go v0.5.4-alpha.1`

Is that what we want or should we just drop it on the floor (maybe log something??) due to the re-write not having a version. 

------------------

Also fair warning I didn't make that many changes to the test but I did split them from the one big test into their own individual ones so that they could be run individually while validating this change.

* Fixes #207 

cc @bhamail / @DarthHater / @aeneasr


zendern:
wrong account will reopen in a minute.

PR #224, opened 2021-03-15
Config via env

bhamail:
Flip the switches required to make Viper read environment variables for configuration values.

cc @bhamail / @DarthHater


zendern (approved):
Looks good. 👍

PR #225, opened 2021-04-13
Let some people know they can try out newer versions, if we know they aren't bad!

DarthHater:
It dawned on me that because `go list -m -json -u all` let's us know the newest version of something, we can run that version past OSS Index and see if it's vulnerable or not. If it's not, let's suggest someone update to it!

* way too many to count right now, putting this up to start getting feedback!

Of note, this is slower, but it's because `go list -m -json -u all` is making requests to I assume `pkg.go.dev` to see what the newest version of a package is

TODO: 
- go through and fix busted tests, add good new ones
- modify how we pass things around such that we know if someone used `dep` or `go mod`, so we can give them the right instructions on how to fix things
- suggest a `go get -u` command to users if the dependency is in the `go.mod` file, and then the `replace` if the dependency is not in their go.mod file?

This builds the path towards a `nancy solve` command where we can go ok, you want us to take care of this? We will give it a shot! That command can work with both `ossindex` and `iq`, because we should get similar data from IQ, knowing what the recommended version of something is.

Screen Shot 2021-04-12 at 10 50 25 PM

cc @bhamail / @DarthHater


go.mod
```diff
@@ -37,10 +37,10 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
 )
 
-// fix vulnerability: CVE-2020-15114 in etcd v3.3.13+incompatible
-replace github.com/coreos/etcd => github.com/coreos/etcd v3.3.24+incompatible
+// // fix vulnerability: CVE-2020-15114 in etcd v3.3.13+incompatible
+// replace github.com/coreos/etcd => github.com/coreos/etcd v3.3.24+incompatible
```
DarthHater, line 41: I have these commented out for easy local testing of vulnerable dependencies

go.mod
```diff
@@ -37,10 +37,10 @@ require (
...
 
-// fix vulnerability: CVE-2021-3121 in github.com/gogo/protobuf v1.2.1
-replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+// // fix vulnerability: CVE-2021-3121 in github.com/gogo/protobuf v1.2.1
+// replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
```
DarthHater, line 44: I have these commented out for easy local testing of vulnerable dependencies

internal/audit/auditlog.go
```diff
@@ -20,51 +20,37 @@ import (
...
 
 // LogResults will given a number of expected results and the results themselves, log the
 // results.
-func LogResults(formatter log.Formatter, packageCount int, coordinates []types.Coordinate, invalidCoordinates []types.Coordinate, exclusions []string) int {
-	vulnerableCount := 0
+func LogResults(formatter log.Formatter, packageCount int, coordinates map[string]ossIndexTypes.Coordinate, invalidCoordinates []ossIndexTypes.Coordinate, vulnerableCoordinates map[string]types.Projects, exclusions []string) int {
 
-	for _, c := range coordinates {
-		c.ExcludeVulnerabilities(exclusions)
+	if exclusions == nil {
```
DarthHater, line 32: I moved this further up because it seemed odd it would make this AFTER we try using exclusions?

internal/audit/auditlogtextformatter.go
```diff
@@ -153,47 +155,54 @@ func groupAndPrint(vulnerable []types.Coordinate, nonVulnerable []types.Coordina
...
-func splitPackages(entries []types.Coordinate) (nonVulnerable []types.Coordinate, vulnerable []types.Coordinate) {
-	for _, v := range entries {
-		if v.IsVulnerable() {
-			vulnerable = append(vulnerable, v)
-		} else {
-			nonVulnerable = append(nonVulnerable, v)
-		}
-	}
-	return
-}
```
DarthHater, line 199: This wasn't even being used, or rather it was, but we already had two lists of data. I removed it because LOL

packages/packages.go
```diff
@@ -44,3 +48,11 @@ func convertGopkgNameToPurl(name string) (rename string) {
 	}
 	return
 }
+
+func GimmeAPurl(name string, version string) string {
```
DarthHater, line 52: I can change this name, but I was laughing as I typed it

gmohre:
```suggestion
func gimmeAPURL(name string, version string) string {

File: internal/audit/auditlogtextformatter.go [hunk 6]
```diff
@@ -76,19 +77,19 @@ func logInvalidSemVerWarning(sb *strings.Builder, noColor bool, quiet bool, inva
 	}
 }
 
-func logVulnerablePackage(sb *strings.Builder, noColor bool, coordinate types.Coordinate) {
+func logVulnerablePackage(sb *strings.Builder, noColor bool, coordinate types.Projects) {
```
bhamail, line 80: I was struggling with the change to `types.Projects`. Made more sense once I saw how the term is used in the `dep` code, e.g.: `dep.Project` - oh yeah, a coordinate really refers to a given "project" that we depend on...Oh the joys of Naming.

internal/cmd/root.go
```diff
@@ -443,21 +441,70 @@ func doStdInAndParse(ossIndex ossindex.IServer) (err error) {
...
+	for k := range coordinates {
+		purls = append(purls, k)
+	}
+
+	ossIndexResponse, err := ossIndex.Audit(purls)
 	if err != nil {
 		return
 	}
 
+	// Wittle down list of audited to vulnerable stuff, so we can work faster
```
bhamail, line 461: typo: Whittle ;)

bhamail (approved):
Seems mostly sane to me. Project=Dep and Coordinate stuff gets confusing, but I don't see a clearer path. "Naming" (pronounce like Seinfeld "Newman").

For later discussion (version 2?): When we find a "clean" newer version, is a "replace" directive the best solution to suggest? If we could determine if the dependency is non-transitive (is direct - not indirect), then maybe suggesting a `go get -u mydep@v` command is better? Probably a lot more trouble, but wanted to ask. In my mind, the problem with adding `replace` directives is they are a form of technical debt - ideally, you always prefer upgrading to new versions of deps instead. With `replace` directives, you need to go back "someday" and determine if the directive can be removed in favor of a newer dependency version.

packages/packages.go
DarthHater: Haha YESSSSS

internal/audit/auditlogtextformatter.go
DarthHater: Yeah I'm not attached to this name TBH. The changes in this PR would merit a new major version of Nancy, so if we want to clean up some naming, let's do it!

DarthHater:
> Seems mostly sane to me. Project=Dep and Coordinate stuff gets confusing, but I don't see a clearer path. "Naming" (pronounce like Seinfeld "Newman").
> 
> For later discussion (version 2?): When we find a "clean" newer version, is a "replace" directive the best solution to suggest? If we could determine if the dependency is non-transitive (is direct - not indirect), then maybe suggesting a `go get -u mydep@v` command is better? Probably a lot more trouble, but wanted to ask. In my mind, the problem with adding `replace` directives is they are a form of technical debt - ideally, you always prefer upgrading to new versions of deps instead. With `replace` directives, you need to go back "someday" and determine if the directive can be removed in favor of a newer dependency version.

I absolutely agree with you. What I was THINKING is we can parse `go.mod` to see if the dependency is in there, and if it is, we can suggest `go get -u <dependency>`, and if not, suggest the replace. It'd be nice if the go API had the ability to let us see what brought something in (which is slightly possible with `go mod graph` but it's not maybe the cleanest way of getting things)

What do you think about changing `Project` to `Dependency` because that might be clearer?

internal/cmd/root.go
DarthHater: WOOD YOU GIVE ME A BREAK!?!

bhamail:
> What do you think about changing `Project` to `Dependency` because that might be clearer?

Yes, I think that change might ease the cognitive load.

internal/audit/csvformatter.go
```diff
@@ -39,21 +39,20 @@ func (f CsvFormatter) Format(entry *Entry) ([]byte, error) {
...
 		invalidEntries := entry.Data["invalid"].([]types.Coordinate)
-		packageCount := entry.Data["num_audited"].(int)
-		numVulnerable := entry.Data["num_vulnerable"].(int)
 		buildVersion := entry.Data["version"].(string)
+		numVulnerable := len(vulnerableEntries.(map[string]interface{}))
+		numPackages := len(auditedEntries)
 
 		var summaryHeader = []string{"Audited Count", "Vulnerable Count", "Build Version"}
 		var invalidHeader = []string{"Count", "Package", "Reason"}
 		var auditedHeader = []string{"Count", "Package", "Is Vulnerable", "Num Vulnerabilities", "Vulnerabilities"}
```
zendern, line 54: Should we also maybe add these "can be updated" packages to the csv output too?? Maybe a column of "Latest non-vulnerable version" or something like that??

internal/audit/auditlog.go
zendern: so the only reason it was down there was b/c if you log it it would not work correctly i believe. So if you passed in nil we wanted it to present to the user as `[]`. Its used in the ExcludeVulnerabilites but range over nil is noop ¯\_(ツ)_/¯ 

https://github.com/sonatype-nexus-community/go-sona-types/blob/b469b94b5ebb013125d69c3524ba5d6f759eede0/ossindex/types/types.go#L103-L107

So really we only do it for presentation. If we can find a better way im cool with it. But moving it up or down doesn't mean much just has to be before the log statement at line 46 hits.

packages/packages.go
```diff
@@ -44,3 +48,11 @@ func convertGopkgNameToPurl(name string) (rename string) {
 	}
 	return
 }
+
+func GimmeAPURL(name string, version string) string {
+	version = strings.Replace(version, "v", "", -1)
+	version = strings.Replace(version, "+incompatible", "", -1)
```
zendern, line 54: Do we need to strip more here??  `+meta`, `-pre`, etc etc.

https://golang.org/ref/mod#versions

DarthHater: I THINK I stripped that version part out for OSS Index, IIRC. I'd have to look further into it.

PR #238, opened 2021-05-28
228 - support for go list -deps

### LINKED ISSUE #228: Suggested `go list -m json all` vulnerability checks swathes of dependencies that never end up in a binary

When checking for vulnerabilities in Golang dependencies using Sonatype's OSS Index via `nancy sleuth`, the README advises to run `go list -json -m all | nancy sleuth [flags]`. However, this will frequently flag up vulnerabilities for test dependencies and dependencies of dependencies that never actually end up in any resulting binary, but are simply tracked by the go.mod/go.sum. These don't show up in `go list -deps` or `go version -m <binary>` and the only way to satisfy the scanner for them is to put arbritrary `replace { ... }` directives to bump their versions even though it has no effect on the binary.

Parsing the json output of `go list -deps -json ./...` should give a more accurate picture of the dependencies that are actually used to make up the project binaries. Currently it is possible to mux that into a `go list -m all` style format by doing something like:

```
go list -deps -json ./... \
        | jq -s 'unique_by(.Module.Path)|.[]|select(has("Module"))|.Module' \
        | nancy sleuth [flags]
```

But I think it would be helpful if the sleuth command itself could understand the `go list -deps -json` format itself and directly parse it.

cc @bhamail / @DarthHater

SirMaster:
This is an initial look into allowing `go list -deps` to be supported, `go list -deps -json` gives us the specific dependencies that will be used at runtime for a binary of go. By allowing a user to use `go list -deps -json ./... | nancy sleuth` we give control to them by allowing them to choose what to check for vulenerabilities.

* Additional support to slurp in the result stream of `go list -deps -json`

* Fixes #228 

cc @bhamail / @DarthHater

**Note** created a draft PR first to confirm code changes first, if all is good will do documentation changes as well in a full PR.


DarthHater:
@dnwe , the infallible @SirMaster has cranked this out, if you wouldn't mind taking a gander.

dnwe:
@SirMaster I gave this a quick test on Go 1.16, but it looks like it's not currently extracting the module deps from the the input json. Building from your branch and giving it the `-deps -json` output nancy reports that it didn't scan any modules

```
$ go list -deps -json ./... | ./nancy sleuth
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Summary                     ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━┫
┃ Audited Dependencies    ┃ 0 ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━┫
┃ Vulnerable Dependencies ┃ 0 ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━┛
```

DarthHater:
Closing in favor of #247

PR #247, opened 2021-09-16
Implement parsing of go list -deps -json ./... and maintain the rest

### LINKED ISSUE #228: Suggested `go list -m json all` vulnerability checks swathes of dependencies that never end up in a binary

DarthHater:
This should resolve #228 , it adds the ability to parse that type of output, and streamlines the code a tiny bit (at the expense of stronger typing, sadly)

For others edification:
- scanning a project with `go list -json -m all` includes every dependency you are using, on paper, and includes info on replace blocks, etc... so this is the recipe
- scanning a project with `go list -json -deps` tells go to let us know about ONLY the dependencies that end up in the final product, so this the cake, what got built with the recipe. This excludes test deps best I can tell, although they are reference in some of the json we end up with (only to the point where we see that a package has test dependencies)

* Adds a test in `parse_test.go`
* Rethinks a few things in `parse.go` around how we parse json, using a `map[string]interface{}` as what we originally parse in to, such that we can just check if keys exist
* Removes dead code
* Updates docs, etc...

* Fixes #228 

cc @bhamail / @DarthHater

A nice bow to @SirMaster who got the groundwork laid, too!


parse/parse.go
```diff
@@ -52,22 +51,46 @@ func GoListAgnostic(stdIn io.Reader) (deps types.ProjectList, err error) {
...
 		}
-		if err != nil {
+		if decodeErr != nil {
+			err = decodeErr
 			break
 		}
 
-		project, err := modToProjectList(mod)
-		if _, ok := err.(*NoVersionError); ok {
+		if module, ok := mod["Module"].(map[string]interface{}); ok {
```
DarthHater, line 66: The easiest way I found to kind of tell if something was of `go list -m all` style versus `go list -deps -json` was this. I'm doing type checks on map keys, basically looking for existence. This lost us some of the nicer json parsing into a struct, but the amount of data we end up using is so minimal this didn't seem too wild. If a `Module` key exists, I know I'm in `go list -deps` town. If it doesn't, I look for `Path`, which is `go list -m all` town. If we find none of that, we just parse like it's a regular list (which we should drop in the future, there's not much of a reason to keep it around).

parse/parse.go
```diff
@@ -52,22 +51,46 @@ func GoListAgnostic(stdIn io.Reader) (deps types.ProjectList, err error) {
...
+			if version, ok := module["Version"].(string); ok {
+				deps.Projects = append(deps.Projects, types.Projects{Version: version, Name: module["Path"].(string)})
```
DarthHater, line 68: I was iffy on doing another type assertion check on `module["Path"]` because I know if `Version` is there then `Path` is but if anyone wants me to go through the extra hoop let me know!

parse/parse.go
bhamail: Neato. The explanation above makes total sense now, while it's on my mind. It feels like adding some code comments around the "why" of these check might avoid head scratching down the road.

BTW, this looks very nice to me: `207` vs `49`

```
$ go list -json -m all | ./nancy sleuth
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Summary                       ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━┫
┃ Audited Dependencies    ┃ 207 ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━━━┫
┃ Vulnerable Dependencies ┃ 0   ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━━┛
(base) MBP-DRollo5:nancy bhamail$ go list -json -deps | ./nancy sleuth 
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Summary                      ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━┫
┃ Audited Dependencies    ┃ 49 ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━━┫
┃ Vulnerable Dependencies ┃ 0  ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━┛
```

parse/parse.go
bhamail: I'm less interested in the extra hoop, but some code comments about the "why" might illuminate.

bhamail (approved):
+1

I'm thinking we should change the docs to suggest the use of `go list -json -deps`.
Would it be helpful if I pushed some doco changes?

DarthHater:
cc @dnwe

PR #248, opened 2021-09-18
show number of ignored vulnerabilities in summary table

### LINKED ISSUE #232: Summary should show the number or list of ignored vulnerabilities
right now if the `.nancy-ignore` file contains anything, then the Nancy just silently ignore them, but let at least print the list of ignores in summary. So the CI/CD engineers can see it in the logs of CI jobs.

cc @bhamail / @DarthHater

MichelKazi:
* Added an exclusion count to auditlog and auditlogtextformatter
* Now in summary table, there is a number of ignored deps highlighted in yellow, kinda like a warning

I'm unsure if y'all would rather it not be in the table though. I also think I could probably just list out the dependencies that are ignored but I am just now re-learning go so baby steps :)

I REALLY miss you all. <3

* Fixes  #232

cc @bhamail / @DarthHater

### DIFF
--- a/internal/audit/auditlog.go
+++ b/internal/audit/auditlog.go
@@ -28,6 +28,7 @@ import (
 // results.
 func LogResults(formatter log.Formatter, packageCount int, coordinates []types.Coordinate, invalidCoordinates []types.Coordinate, exclusions []string) int {
 	vulnerableCount := 0
+	exclusionCount := 0
 
 	for _, c := range coordinates {
 		c.ExcludeVulnerabilities(exclusions)
@@ -35,9 +36,16 @@ func LogResults(formatter log.Formatter, packageCount int, coordinates []types.C
 
 	var auditedCoordinates []types.Coordinate
 	var vulnerableCoordinates []types.Coordinate
+	var excludedVulnerabilities []types.Vulnerability
 
 	for i := 0; i < len(coordinates); i++ {
 		coordinate := coordinates[i]
+		for _, v := range coordinate.Vulnerabilities {
+			if v.Excluded {
+				exclusionCount++
+				excludedVulnerabilities = append(excludedVulnerabilities, v)
+			}
+		}
 		if coordinate.IsVulnerable() {
 			vulnerableCount++
 			vulnerableCoordinates = append(vulnerableCoordinates, coordinate)
@@ -54,14 +62,17 @@ func LogResults(formatter log.Formatter, packageCount int, coordinates []types.C
 	if vulnerableCoordinates == nil {
 		vulnerableCoordinates = make([]types.Coordinate, 0)
 	}
+
 	log.SetFormatter(formatter)
 	log.SetOutput(os.Stdout)
 	log.WithFields(log.Fields{
 		"exclusions":     exclusions,
 		"num_audited":    packageCount,
 		"num_vulnerable": vulnerableCount,
+		"num_exclusions": exclusionCount,
 		"audited":        auditedCoordinates,
 		"vulnerable":     vulnerableCoordinates,
+		"excluded":       excludedVulnerabilities,
 		"invalid":        invalidCoordinates,
 		"version":        buildversion.BuildVersion,
 	}).Info("")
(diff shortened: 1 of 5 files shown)


MichelKazi:
Feels like a brute force leetcode solution to nest a for loop and check for vulnerabilities here

maybe I can add an Exclusions field here https://github.com/sonatype-nexus-community/go-sona-types/blob/main/ossindex/types/types.go and just get the length of that or something

MichelKazi:
@svilgelm 
image

Let me know what you think

SVilgelm:
Looks good 👍

SVilgelm (approved):

zendern:
PR changes look good to me. I think we could use a test to validate it does in fact get printed out.

See here : 
https://github.com/sonatype-nexus-community/nancy/blob/main/internal/audit/auditlogtextformatter_test.go

And if you are up for it we might as well add it to the csv formatter too. 
https://github.com/sonatype-nexus-community/nancy/blob/main/internal/audit/csvformatter.go

json output it already happens for free as we just serialize the object that gets passed in.

internal/audit/auditlog.go
```diff
@@ -28,16 +28,24 @@ import (
...
 
 	var auditedCoordinates []types.Coordinate
 	var vulnerableCoordinates []types.Coordinate
+	var excludedVulnerabilities []types.Vulnerability
 
 	for i := 0; i < len(coordinates); i++ {
 		coordinate := coordinates[i]
+		for _, v := range coordinate.Vulnerabilities {
+			if v.Excluded {
+				exclusionCount++
```
DarthHater, line 45: Minor: You can likely just do `len(excludedVulnerabilities)` to skip having another variable for this

internal/audit/auditlogtextformatter.go
```diff
@@ -153,14 +153,17 @@ func groupAndPrint(vulnerable []types.Coordinate, nonVulnerable []types.Coordina
 func (f AuditLogTextFormatter) Format(entry *Entry) ([]byte, error) {
 	auditedEntries := entry.Data["audited"]
 	invalidEntries := entry.Data["invalid"]
+	excludedEntries := entry.Data["excluded"]
 	packageCount := entry.Data["num_audited"]
 	numVulnerable := entry.Data["num_vulnerable"]
+	numExcluded := entry.Data["num_exclusions"]
 	buildVersion := entry.Data["version"]
-	if auditedEntries != nil && invalidEntries != nil && packageCount != nil && numVulnerable != nil && buildVersion != nil {
+	if auditedEntries != nil && invalidEntries != nil && excludedEntries != nil && packageCount != nil && numVulnerable != nil && numExcluded != nil && buildVersion != nil {
```
DarthHater, line 161: This is just an observation but this is UGLLLLY as it grows.

You might create a func like:

```
func isNil(params ...interface{}) bool {
  foundNil := false
  for _, v := range params {
     if v == nil || reflect.ValueOf(v).IsNil() {
       foundNil = true
     }
  }
  return foundNil
}
```

Then hypothetically you can just do `isNil(auditedEntries, invalidEntries, excludedEntries, packageCount, numVulnerable, numExcluded, buildVersion)`

This is pseudo code so no idea if it compiles, just an idea so we don't end up with a large finger bug some day :)

internal/audit/auditlogtextformatter.go
```diff
@@ -179,6 +182,10 @@ func (f AuditLogTextFormatter) Format(entry *Entry) ([]byte, error) {
 		t.AppendRow([]interface{}{"Audited Dependencies", strconv.Itoa(packageCount)})
 		t.AppendSeparator()
 		t.AppendRow([]interface{}{"Vulnerable Dependencies", au.Bold(au.Red(strconv.Itoa(numVulnerable)))})
+    if numExcluded > 0 {
```
bhamail, line 185: super minor: formatting went wonky here. oh wait, who was using tabs previously!?!? Let the holy wars begin.

bhamail (approved):
+1

DarthHater (approved):

PR #249, opened 2021-09-22
Allow for base path of DB cache to be set

### LINKED ISSUE #216: Configurable cache directory

Set a unique cache directory

Caching results between runs

I believe this will need some changes: https://github.com/sonatype-nexus-community/go-sona-types/blob/a618cf1b0a4b00832ff05cd83f16114d5bc43c48/ossindex/internal/cache/cache.go#L67

* **Anything else?**

cc @bhamail / @DarthHater

DarthHater:
This should address #216 

* Upgrades `go-sona-types` to `0.1.2`
* Adds a `db-cache-path` flag
* Sets `db-cache-path` when setting up OSS Index
* Updates README, etc...

* Fixes #216 

cc @bhamail / @DarthHater


bhamail (approved):
LGTM. As soon as CI is happy, will push release buttons again.

PR #250, opened 2021-09-23
function to validate formatter fields for cleaner code

MichelKazi:
* prevents super long one-line if statement and just passes Entry.Data fields into a validator

cc @bhamail / @DarthHater

### DIFF
--- a/internal/audit/auditlog.go
+++ b/internal/audit/auditlog.go
@@ -24,11 +24,19 @@ import (
 	"github.com/sonatype-nexus-community/nancy/buildversion"
 )
 
+func isEntryValid(params ...interface{}) bool {
+	for _, v := range params {
+		if v == nil {
+			return false
+		}
+	}
+	return true
+}
+
 // LogResults will given a number of expected results and the results themselves, log the
 // results.
 func LogResults(formatter log.Formatter, packageCount int, coordinates []types.Coordinate, invalidCoordinates []types.Coordinate, exclusions []string) int {
 	vulnerableCount := 0
-	exclusionCount := 0
 
 	for _, c := range coordinates {
 		c.ExcludeVulnerabilities(exclusions)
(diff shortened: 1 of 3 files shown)


MichelKazi:
Also have vim running GoFmt properly before exit so it's formatted to however y'all like it <3

internal/audit/auditlog.go
```diff
@@ -24,11 +24,19 @@ import (
...
 
+func isEntryValid(params ...interface{}) bool {
+	for _, v := range params {
+		if v == nil {
+			return false
+		}
+	}
+	return true
+}
+
```
DarthHater, line 35: This will work fine for what we are doing, but if you are curious you can take a look at this for some cases where it WONT work: https://mangatmodi.medium.com/go-check-nil-interface-the-right-way-d142776edef1

Also, this sent me on a quick lil check on if golang likes yoda conditions, and it does not! https://staticcheck.io/docs/checks#ST1017

internal/audit/auditlog.go
```diff
@@ -63,6 +70,8 @@ func LogResults(formatter log.Formatter, packageCount int, coordinates []types.C
 		vulnerableCoordinates = make([]types.Coordinate, 0)
 	}
 
+	exclusionCount := len(excludedVulnerabilities)
```
DarthHater, line 73: 🎉

DarthHater:
@bhamail is there a reason CI hasn't run on this?

bhamail:
Hm, can't figure out why CI is not triggering. I tried "redelivering" some of the webhook events, but no joy. 
I'll post on the CircleCI forums and see what I can find.

internal/audit/auditlog.go
MichelKazi: Would you like me to update this? I can also move this function out to somewhere where we can use it everywhere instead of just audit

internal/audit/auditlog.go
MichelKazi: I’m a die hard 2 spaces fan :)
But consistency takes priority

bhamail (approved):
LGTM. No word from CircleCI about why the build ain't buildin'. Gonna merge this puppy and see what happens. duck and cover.

PR #267, opened 2022-06-02
fix #264: restore v version prefix

### LINKED ISSUE #264: Update version string to have "v" prefix expected by OSSIndex
Recent changes to OSSIndex now expect golang component versions to be prefixed with the letter "v". We should update to follow this pattern. Might only need to change go-sona-types lib?

cc @bhamail / @DarthHater

bhamail:
OSS Index used to require the version for golang components to not have a "v" prefix. This has changed, so now both OSS Index and IQ are happy with golang components who's version string is prefixed with a "v".

This PR removes the code that removed the "v".

* Fixes #264

cc @bhamail / @DarthHater


DarthHater (approved):
Nice sensible changes that seem to remove something I really hated doing to begin with.

PR #269, opened 2022-06-14
Support additional exclude vulnerability files

uvegla:
This change adds the possibility to split `.nancy-ignore` files into multiple ones in a non-breaking way. The "root" file is still kept as `.nancy-ignore` or can be overwritten with the `-x` / `--exclude-vulnerability-file` flag but on top of that a list of additional files can be set via the `-a` / `--additional-exclude-vulnerability-files` flag.

The most common use case for using multiple files is when an organisation wants centrally control some of the common known vulnerabilities across hundreds of repositories by generating a file. It makes sense when there are common 3rd party libraries used in upstream dependencies that are either not updated anymore and the organisation cannot change the usage of them in a short period of time, but it takes a short migration for example. By generating am ignore file via automated means controlled by a central force it can save the headache for the rest of the organisation to handle these one-by-one.

However there can still be known vulnerabilities a team wants to ignore for their own repository that is specific to only that one, so the build pipeline for an organisation could look something like:

```shell
go list -json -deps | nancy sleuth -a .nancy-ignore.local
``` 

I chose to add an additional flag to keep backwards compatibility when someone uses the `-x` / `--exclude-vulnerability-file`. An other approach could be to change that flag to a list instead and rename it to `-x` / `--exclude-vulnerability-files`. Because of this it is still possible to control the root ignore file via that flag and do something like the below example.

```shell
go list -json -deps | nancy sleuth -x .nancy-ignore.generated -a .nancy-ignore.local
```

- [x] If there are changes to user behavior in general, please make sure to update the docs, as well

cc @bhamail / @DarthHater


uvegla:
@bhamail @DarthHater Can I please get your feedback / review on this change? Thanks in advance! 🙇🏻

bhamail (approved):
+1   Nice test coverage! Thanks!!! I plan to merge this puppy pronto. Please holler if you object to either of my changes - happy to revert them if needed.

PR #277, opened 2024-02-26
Update README.md

orsenthil:
Explain the relationship and difference with govulncheck

cc @bhamail / @DarthHater


bhamail (approved):
Thanks for contributing this! LGTM. +1

PR #306, opened 2026-06-17
feat: set maximum go list input size via CLI flag

### LINKED ISSUE #305: 10Mb Input size limits on stdin prevents CLI usage on many repositories
[This change](https://github.com/sonatype-nexus-community/nancy/issues/291) broke our CI workflows. The 10Mb limit feels too tight, specially since it can't be configured when invoking the CLI.

We use `nancy` in our CI workflows for several different repositories with different sizes, and mid size projects go over the 10Mb limit pretty easily.

I see two options to improve the situation:
- option1: a significantly bigger limit is set
- option2: offer a new CLI flag to configure the limit so that users can choose

stone-z:
https://github.com/sonatype-nexus-community/nancy/pull/298 introduced a hard-coded cap on the input size `nancy` would accept from `go list`. This breaks CI larger projects where the output size exceeds the limit.

* increases the default limit to 100MB
* makes the cap configurable using a new command like flag `max-go-list-input-size`

This changes the `GoListAgnostic` function signature, which might be considered breaking. If the maintainers would prefer, I'm open to changing the function to accept a config object that leaves it open to future options without touching the signature again, but it seemed like overkill at this point.

* Fixes https://github.com/sonatype-nexus-community/nancy/issues/305

cc @sonatype-nexus-community/nancy

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@
 - [Sonatype Guide Options](#sonatype-guide-options)
   - [Using a custom server URL](#using-a-custom-server-url)
   - [Loud mode](#loud-mode)
+  - [Maximum input size](#maximum-input-size)
   - [Exclude vulnerabilities](#exclude-vulnerabilities)
   - [Output](#output)
 - [Sonatype Lifecycle Options](#sonatype-lifecycle-options)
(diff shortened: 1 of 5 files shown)


stone-z:
Signed and signed-off now :heavy_check_mark: 

(btw, the contribution guide [points to](https://github.com/sonatype-nexus-community/nancy/blob/b23b5eb9fac69a345cbaaf5a982d357afdd1f2ad/CONTRIBUTING.md?plain=1#L19) an [old/incorrect CLA page](https://sonatypecla.herokuapp.com/sign-cla))

madpah (approved):

madpah:
Thanks for the PR @stone-z 🥳

